PR #25, opened 2019-12-22
[ExchangeManager] Manage stop properly

Herklos:
requires #24


octobot_trading/exchanges/exchange_manager.py
```diff
@@ -109,42 +115,51 @@ def reset_exchange_personal_data(self):
...
+            if self.check_web_socket_config(self.exchange.name):
+                await self._search_and_create_websocket(AbstractWebsocket)
+
+    async def _create_simulated_exchange(self):
+        self.exchange = ExchangeSimulator(self.config, self.exchange_type, self, self.backtesting_files)
+        await self.exchange.initialize()
+        self.exchange_config.set_config_traded_pairs()
+        await self._create_exchange_channels()
+
+    async def _init_simulated_exchange(self):
```
GuillaumeDSM: No need for this and the 3 previous methods in pxd ?

Herklos: It's not possible to declare async methods in pxd.

PR #26, opened 2019-12-24
[PriceManager] Move mark price calculation method

Herklos:
requires #25


octobot_trading/data_manager/prices_manager.py
```diff
@@ -35,9 +35,13 @@ def set_mark_price(self, mark_price):
...
+    async def get_mark_price(self, timeout=MARK_PRICE_TIMEOUT):
+        await wait_for(self.prices_initialized_event.wait(), timeout)
         return self.mark_price
 
     def __reset_prices(self):
         self.mark_price = 0
+
+    @staticmethod
+    def calculate_mark_price_from_recent_trade_prices(recent_trade_prices):
+        return sum(recent_trade_prices) / len(recent_trade_prices)
```
GuillaumeDSM: :+1:  actually since this is just an average compuration, maybe we should use an utility method for this, no price specific computation is being done here to justify its place in prices_manager

Herklos: Yes but I think we'll have to improve the calculation method, so it will be easier to modify with this implementation.

PR #34, opened 2019-12-28
[CandleManager] Fix starting candle and unwanted data 

Herklos:


octobot_trading/data_manager/candles_manager.py
```diff
@@ -171,7 +170,7 @@ def __inc_candle_index(self):
 
     def __extract_limited_data(self, data, limit=-1, max_limit=-1):
         if limit == -1:
-            return data
+            return data[~np.isnan(data)]
 
         if max_limit == -1:
             return data[-min(limit, len(data)):]
```
GuillaumeDSM: Don't we also need to filter np.isnan here and in else ?

Herklos: I don't think so, because in that case you ask for a specific array size so I think it's better to return the array you ask with np.nan values in it.

Herklos: I other words your prefer to return when the requested slice is full of np.nan : 
- []
- [nan, nan, nan...]

Honestly I don't really know which one is better.

GuillaumeDSM: I think the risk here with fill value (not just in if and else but in the whole ```__extract_limited_data```) is if an exchange is saying 'nan' to a value but not all of them, it will result in invalid candled (values will be shifted only in the list with the added nan)

=> I think we should either make sure that nan is never added except at init or use an impossible candle info (like -2 or something similar) => from that we should in my opinion never return any data containing this value, even if the limit asked is higher that the current available data length.
My main point is that if we allow users of the candle manager to receive invalid (and internal list filling value) that is either false or unusable, we will have to filter this value in every candle user because we can't use it anyway.
Maybe we should add an optionnal param to tell not to filter if a user really wants untouched data but since it's only initialization filling info, I hardly see anything using it directly

Herklos: Ok I understand your point of view. It seems logic to have the same behavior and to exclude the filtering of exchange missing data (with the same filling value aka np.nan).

But the problem is that we have to choose the correct filling value. I don't think that something negative as -1 or -2 is a good idea. It's the best solution for now, but in the future if we want to add index chart candle or something like this that exists in stock market. We can have such values in candles.

I'm not sure if exchange really give an np.nan for incorrect values. I think we have to search for a compatibility between  numpy additionnal values (nan or something else) and ccxt exchange candle incorrect values.

Herklos: @GuillaumeDSM

PR #27, opened 2019-12-24
[Traders] Fix legacy migration

Herklos:
requires #25

PR #53, opened 2019-12-30
[Trades] Implement factory

Herklos:
Requires #27


octobot_trading/data/trade.py
```diff
@@ -14,23 +14,54 @@
...
+        self.total_cost = order.total_cost
+        self.trade_type = order.order_type
+        self.status = order.status
+        self.fee = order.fee
+        self.trade_id = order.order_id
+        self.side = order.side
+        self.creation_time = order.creation_time if order.creation_time > 0 else creation_time
+        self.canceled_time = order.canceled_time if order.canceled_time > 0 else canceled_time
+        self.executed_time = order.executed_time if order.executed_time > 0 else executed_time
+        self.symbol = order.symbol
```
GuillaumeDSM: It looks like ```taker_or_maker```, ```created_last_price``` and ```origin_stop_price``` are not updated. Not sure if ```created_last_price``` and ```origin_stop_price``` are really relevant but ```taker_or_maker``` should be updated

Herklos: Right !

octobot_trading/data/trade.pxd
```diff
@@ -30,20 +30,19 @@ cdef class Trade:
     cdef public object side # TradeOrderSide
     cdef public object status # OrderStatus
     cdef public object trade_type # TraderOrderType
+    cdef public object taker_or_maker # ExchangeConstantsMarketPropertyColumns
```
GuillaumeDSM: Is it not a str ?

Herklos: Yes it's a string, I used the order.pxd definition of it but it's also defined as an object. i will modify both.

octobot_trading/traders/trader.py
```diff
@@ -242,6 +247,11 @@ def set_risk(self, risk):
         if cancel:
             await self.cancel_order(order)
 
+            self.exchange_manager.exchange_personal_data.trades_manager.upsert_trade_instance(
+                create_trade_from_order(order,
+                                        close_status=OrderStatus.CANCELED,
+                                        canceled_time=time.time()))
```
GuillaumeDSM: 👍

PR #71, opened 2020-01-14
[ExchangeChannel] Migrate form exchange name to exchange id

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Evaluators/pull/25 to be approved

### DIFF
--- a/octobot_trading/exchanges/exchange_manager.py
+++ b/octobot_trading/exchanges/exchange_manager.py
@@ -63,7 +63,10 @@ class ExchangeManager(Initializable):
         self.is_backtesting = is_backtesting
         self.is_simulated = is_simulated
         self.is_collecting = is_collecting
+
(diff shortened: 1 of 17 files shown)


Herklos:
How do you imagine to do arbitrage with such implementation ?

We can request a specific exchange manager from its name.

octobot_trading/exchanges/exchange_manager.py
```diff
@@ -87,12 +90,13 @@ def __init__(self, config, exchange_class_string,
         if self.exchange is not None:
             await self.exchange.stop()
             Exchanges.instance().del_exchange(self.exchange.name, self.id)
-            await self.stop_exchange_channels()
+            if not self.exchange_only:
```
Herklos: :+1:

PR #75, opened 2020-01-18
[ExchangeConfiguration] handle matrix id

GuillaumeDSM:


octobot_trading/api/exchange.py
```diff
@@ -63,6 +65,13 @@ def get_exchange_manager_id(exchange_manager) -> str:
     return exchange_manager.id
 
 
+def get_exchange_id_from_matrix_id(exchange_name, matrix_id) -> str:
+    for exchange_configuration in get_exchange_configurations_from_exchange_name(exchange_name).values():
+        if exchange_configuration.matrix_id == matrix_id:
+            return get_exchange_manager_id(exchange_configuration.exchange_manager)
+    return ""
```
Herklos: None maybe ?

GuillaumeDSM: yes, better like this

PR #86, opened 2020-01-23
[Backtesting] update orders management and price channel

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Backtesting/pull/15
requires https://github.com/Drakkar-Software/OctoBot-Backtesting/pull/17

PR #87, opened 2020-01-25
[Backtesting] Init backtesting data and fix issues

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Trading/pull/86
requires https://github.com/Drakkar-Software/OctoBot-Trading/pull/85


octobot_trading/producers/ohlcv_updater.py
```diff
@@ -31,27 +31,42 @@ class OHLCVUpdater(OHLCVProducer):
...
+
     def __init__(self, channel):
         super().__init__(channel)
         self.tasks = []
         self.is_initialized = False
 
+        self.ohlcv_initialized_event = asyncio.Event()
+
+    async def wait_for_initialization(self):
+        await asyncio.wait_for(self.ohlcv_initialized_event.wait(), self.OHLCV_INITIALIZATION_TIMEOUT)
```
Herklos: Do you think we can accept to update candle even if the initialization process timeout ?

octobot_trading/exchanges/exchange_simulator.py
```diff
@@ -61,14 +64,22 @@ def __init__(self, config, exchange_type, exchange_manager, backtesting_data_fil
...
 
     async def create_backtesting_exchange_producers(self):
         for importer in self.exchange_importers:
+            available_data_types = get_available_data_types(importer)
             for updater in UNAUTHENTICATED_UPDATER_SIMULATOR_PRODUCERS:
-                await updater(get_trading_chan(updater.CHANNEL_NAME, self.exchange_manager.id), importer).run()
+                if updater not in SIMULATOR_PRODUCERS_TO_DATA_TYPE \
+                        or any(required_data_type in available_data_types
+                               for required_data_type in SIMULATOR_PRODUCERS_TO_DATA_TYPE[updater]):
+                    if updater == MarkPriceUpdater:
```
Herklos: Why not creating a MarkPriceUpdaterSimulator ?

octobot_trading/data/order.py
```diff
@@ -251,10 +258,10 @@ def get_profitability(self):
             await self.cancel_from_exchange()
 
     def generate_executed_time(self, simulated_time=False):
-        if not simulated_time or not self.last_prices:  # TODO
+        if not simulated_time:  # TODO
             return time.time()
         else:
-            return self.last_prices[-1][ECOC.TIMESTAMP.value]
+            return get_backtesting_current_time(self.exchange_manager.exchange.backtesting)
```
Herklos: :+1:

octobot_trading/exchanges/exchange_simulator.py
GuillaumeDSM: Because it doesn't exist and would be a clone of MarkPriceUpdater, MarkPriceUpdater is a recent trades consumer

octobot_trading/producers/ohlcv_updater.py
GuillaumeDSM: If it timeout, it means we have no historical ohlcv, or it's extremely slow to fetch. I think in this case it's better to keep going with backtesting, if there are more problems with ohlcv fetching, we will now it soon enough. I don't think it's work cancelling the whole backtesting though: if subsequent calls work then backtesting should keep going

Herklos: But here is ohlcv_updater, it's not only a backtesting updater. You should not think backtesting only.

octobot_trading/exchanges/exchange_simulator.py
Herklos: I think it's better to have `MarkPriceUpdaterSimulator` that only herit from `MarkPriceUpdater` but with an `importer` attribute than having a ` if updater == MarkPriceUpdater` here.

octobot_trading/producers/ohlcv_updater.py
GuillaumeDSM: For now this function is only used in backtesting (not necessary for live data), however in case we can to know if ohlcv are initialized in live data, it might make a difference, i'll add an optionnal timeout param.

octobot_trading/producers/simulator/recent_trade_updater_simulator.py
```diff
@@ -46,17 +54,45 @@ def __init__(self, channel, importer):
...
-            self.logger.warning(f"Not enough data : {e}")
+            self.logger.warning(f"Not enough data : {e} will use ohlcv data to simulate recent trades.")
             await self.pause()
             await self.stop()
-        except IndexError as e:
-            self.logger.warning(f"Failed to access recent_trades_data : {e}")
+
+    async def _register_on_recent_trades(self):
+        ohlcv_chan = get_exchange_chan(OHLCV_CHANNEL, self.channel.exchange_manager.id)
+        for producer in ohlcv_chan.get_producers():
```
Herklos: Why do you need to wait that ohlcv producer to be ready ?

GuillaumeDSM: Yes, otherwise it will get called by historical candles push during initialization, which can lead in invalid values

octobot_trading/producers/ohlcv_updater.py
Herklos: Ok but if this method only concerns backtesting, why it has to be implemented in `ohlcv_updater.py` and not just in `ohlcv_updater_simulator.py` ?

octobot_trading/producers/simulator/recent_trade_updater_simulator.py
Herklos: Oh yes ok I see, we have to have the candles related to the current backtesting time to generate recent trades :+1:

Herklos: Can you add some comments in the code to explain this please ? Thanks

octobot_trading/producers/ohlcv_updater.py
GuillaumeDSM: It's called by ```recent_trade_updater_simulator``` as a return value of ```ohlcv_chan.get_producers()```, we technically car put it in ```ohlcv_updater_simulator.py``` but if we use ```recent_trade_updater_simulator``` with another updater than ```ohlcv_updater_simulator.py```, it will crash. We can consider it acceptable though, what do you think ?

Herklos: I understand, maybe the cleaner way is to have an abstract updater simulatr with an abstract wait_for_initialization ? That prevents a simulator to not have wait_for_initialization method. What do you think ?

GuillaumeDSM: I think abstracting it is a good idea, we can make ```wait_for_initialization``` raise an NotImplemented exception on ohlcv_updater.py and have it implemented on ohlcv_updater_simulator.py. This way if we want to implement it later we can and it's explicit. I think it's better than creating a new abstract class that will result in multiple inheritence and won't be cythonizable

Herklos: :+1:

octobot_trading/exchanges/exchange_simulator.py
```diff
@@ -62,16 +61,18 @@ def __init__(self, config, exchange_type, exchange_manager, backtesting_data_fil
...
             maximum_timestamp=max(timestamps)[1])
 
     async def create_backtesting_exchange_producers(self):
         for importer in self.exchange_importers:
+            available_data_types = get_available_data_types(importer)
             for updater in UNAUTHENTICATED_UPDATER_SIMULATOR_PRODUCERS:
-                if updater == MarkPriceUpdater:
-                    await updater(get_trading_chan(updater.CHANNEL_NAME, self.exchange_manager.id)).run()
-                else:
+                if updater not in SIMULATOR_PRODUCERS_TO_DATA_TYPE \
```
Herklos: I understand what you want to do but it's very complex...

octobot_trading/producers/ohlcv_updater.py
```diff
@@ -31,27 +31,43 @@ class OHLCVUpdater(OHLCVProducer):
...
+            await self._initialize()
         self.tasks = [
             asyncio.create_task(self._candle_callback(time_frame, pair))
             for time_frame in self.channel.exchange_manager.exchange_config.traded_time_frames
             for pair in self.channel.exchange_manager.exchange_config.traded_symbol_pairs]
 
+    async def wait_for_initialization(self, timeout=OHLCV_INITIALIZATION_TIMEOUT):
+        raise NotImplementedError("wait_for_initialization is not implemented yet")
+
+    async def _initialize(self):
```
Herklos: :+1:

octobot_trading/producers/prices_updater.pxd
```diff
@@ -15,8 +15,7 @@
 #  You should have received a copy of the GNU Lesser General Public
 #  License along with this library.
 
-from octobot_trading.channels.trades cimport TradesProducer
-from octobot_trading.exchanges.data.exchange_symbol_data cimport ExchangeSymbolData
+from octobot_trading.channels.price cimport MarkPriceProducer
 
-cdef class TradesUpdater(TradesProducer):
+cdef class MarkPriceUpdater(MarkPriceProducer):
```
Herklos: :+1:

octobot_trading/producers/simulator/ohlcv_updater_simulator.py
```diff
@@ -69,3 +76,21 @@ def __init__(self, channel, importer):
         if self.time_consumer is None and not self.channel.is_paused:
             self.time_consumer = await get_chan(OctoBotBacktestingChannelsName.TIME_CHANNEL.value).new_consumer(
                 self.handle_timestamp)
+
+    async def _initialize_candles(self, time_frame, pair):
+        # fetch history
+        ohlcv_data = None
+        try:
+            ohlcv_data: list = await self.exchange_data_importer.get_ohlcv_from_timestamps(
```
Herklos: :+1:

octobot_trading/producers/simulator/orders_updater_simulator.py
```diff
@@ -33,9 +33,11 @@ class OpenOrdersUpdaterSimulator(OpenOrdersUpdater):
     def __init__(self, channel):
         super().__init__(channel)
         self.exchange_manager = None
+        self.simulate_time = False
 
     async def start(self):
         self.exchange_manager = self.channel.exchange_manager
+        self.simulate_time = self.exchange_manager.is_simulated and self.exchange_manager.is_backtesting
```
Herklos: I think `is_simulated_time` should be a better name, what do you think ?

setup.py
```diff
@@ -61,6 +61,7 @@ def build_ext(*args, **kwargs):
                  "octobot_trading.producers.simulator.kline_updater_simulator",
                  "octobot_trading.producers.simulator.orders_updater_simulator",
                  "octobot_trading.producers.simulator.positions_updater_simulator",
+                 "octobot_trading.producers.simulator.price_updater_simulator",
```
Herklos: :+1:

octobot_trading/producers/simulator/recent_trade_updater_simulator.py
```diff
@@ -83,5 +90,11 @@ def _generate_recent_trade(timestamp, price):
...
         if self.time_consumer is None and not self.channel.is_paused:
-            self.time_consumer = await get_chan(OctoBotBacktestingChannelsName.TIME_CHANNEL.value).new_consumer(
-                self.handle_timestamp)
+            if ExchangeDataTables.RECENT_TRADES in get_available_data_types(self.exchange_data_importer):
+                self.time_consumer = await get_chan(OctoBotBacktestingChannelsName.TIME_CHANNEL.value).new_consumer(
+                    self.handle_timestamp)
+            else:
+                # asyncio.shield to avoid auto-cancel (if error in other tasks that will exit main asyncio.gather)
+                # resulting in failure to register as consumer
+                await asyncio.shield(self._register_on_recent_trades())
```
Herklos: :+1:

octobot_trading/exchanges/exchange_manager.py
```diff
@@ -158,6 +159,7 @@ def reset_exchange_personal_data(self):
         try:
             await self.exchange.modify_channels()
             await self.exchange.create_backtesting_exchange_producers()
+            await start_backtesting(self.exchange.backtesting)
```
Herklos: Why this is not made in `ExchangeSimulator` ? Maybe an `ExchangeSimulator.start()` or `.start_backtesting()` ?

octobot_trading/exchanges/exchange_manager.py
```diff
@@ -16,6 +16,7 @@
 import uuid
 from copy import copy
 
+from octobot_backtesting.api.backtesting import start_backtesting
```
Herklos: I don't think the backtesting API should be call in ExchangeManager. I think ExchangeSimulator is the right class to call Backtesting APIs.

octobot_trading/producers/simulator/ohlcv_updater_simulator.py
```diff
@@ -30,13 +28,22 @@ def __init__(self, channel, importer):
...
         self.exchange_name = self.channel.exchange_manager.exchange_name
 
+        self.initial_timestamp = get_backtesting_current_time(self.channel.exchange_manager.exchange.backtesting)
         self.last_timestamp_pushed = 0
         self.time_consumer = None
 
     async def start(self):
+        if not self.is_initialized:
+            await self._initialize()
+        self.is_initialized = True
```
Herklos: `self._initialize()` already put `self.is_initialized` to `True`

octobot_trading/data/portfolio.py
```diff
@@ -49,7 +50,7 @@ def __init__(self, exchange_name, is_simulated=False):
     async def copy(self):
         pf: Portfolio = Portfolio(self.exchange_name, self.is_simulated)
         await pf.initialize()
-        pf.portfolio: dict = self.portfolio.copy()
+        pf.portfolio: dict = deepcopy(self.portfolio)
```
Herklos: :+1:

octobot_trading/producers/simulator/ohlcv_updater_simulator.py
```diff
@@ -13,14 +13,12 @@
 #
 #  You should have received a copy of the GNU Lesser General Public
 #  License along with this library.
-import json
+import asyncio
 
+from octobot_backtesting.api.backtesting import get_backtesting_current_time
```
Herklos: Maybe `.exchange.get_exchange_current_time()` now ?

octobot_trading/producers/simulator/ticker_updater_simulator.py
```diff
@@ -49,11 +56,37 @@ def __init__(self, channel, importer):
...
+    async def _ticker_from_ohlcv_callback(self, exchange: str, exchange_id: str, symbol: str, time_frame, candle):
+        if candle:
+            last_candle_timestamp = candle[PriceIndexes.IND_PRICE_TIME.value]
+            if last_candle_timestamp > self.last_timestamp_pushed:
+                self.last_timestamp_pushed = last_candle_timestamp
+                ticker = self._generate_ticker_from_candle(candle, symbol, last_candle_timestamp)
+                await self.push(symbol, ticker)
+
+    @staticmethod
+    def _generate_ticker_from_candle(candle, symbol, last_candle_timestamp):
```
Herklos: :+1:

PR #88, opened 2020-01-26
Update APIs for backtesting

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Trading/pull/87


octobot_trading/api/exchange.py
```diff
@@ -13,6 +13,7 @@
 #
 #  You should have received a copy of the GNU Lesser General Public
 #  License along with this library.
+from octobot_backtesting.backtesting import Backtesting
```
Herklos: I think this is a bad idea.

PR #89, opened 2020-01-28
[Backtesting] Run backtesting on required timeframes only 

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Trading/pull/87
requires https://github.com/Drakkar-Software/OctoBot-Backtesting/pull/20

### DIFF
--- a/octobot_trading/exchanges/data/exchange_config_data.py
+++ b/octobot_trading/exchanges/data/exchange_config_data.py
@@ -106,8 +106,7 @@ class ExchangeConfig(Initializable):
             if self.exchange_manager.time_frame_exists(time_frame.value):
                 self.traded_time_frames.append(time_frame)
         # add shortest timeframe for realtime evaluators
-        client_shortest_time_frame = find_min_time_frame(
-            self.exchange_manager.client_time_frames[CONFIG_WILDCARD], MIN_EVAL_TIME_FRAME)
+        client_shortest_time_frame = find_min_time_frame(self.exchange_manager.client_time_frames, MIN_EVAL_TIME_FRAME)
         if client_shortest_time_frame not in self.traded_time_frames:
             self.traded_time_frames.append(client_shortest_time_frame)
 
--- a/octobot_trading/exchanges/data/exchange_config_data.py
+++ b/octobot_trading/exchanges/data/exchange_config_data.py
@@ -105,10 +105,11 @@ class ExchangeConfig(Initializable):
         for time_frame in get_config_time_frame(self.config):
             if self.exchange_manager.time_frame_exists(time_frame.value):
                 self.traded_time_frames.append(time_frame)
-        # add shortest timeframe for realtime evaluators
-        client_shortest_time_frame = find_min_time_frame(self.exchange_manager.client_time_frames, MIN_EVAL_TIME_FRAME)
-        if client_shortest_time_frame not in self.traded_time_frames:
-            self.traded_time_frames.append(client_shortest_time_frame)
+        if not self.exchange_manager.is_backtesting:
+            # add shortest timeframe for realtime evaluators
+            client_shortest_time_frame = find_min_time_frame(self.exchange_manager.client_time_frames, MIN_EVAL_TIME_FRAME)
+            if client_shortest_time_frame not in self.traded_time_frames:
+                self.traded_time_frames.append(client_shortest_time_frame)
 
         self.traded_time_frames = sort_time_frames(self.traded_time_frames, reverse=True)
 
--- a/octobot_trading/exchanges/exchange_manager.py
+++ b/octobot_trading/exchanges/exchange_manager.py
@@ -20,7 +20,7 @@ from octobot_channels.util.channel_creator import create_all_subclasses_channel
 from octobot_websockets.constants import CONFIG_EXCHANGE_WEB_SOCKET
 
 from octobot_commons.config_util import has_invalid_default_config_value
-from octobot_commons.constants import CONFIG_ENABLED_OPTION, CONFIG_WILDCARD
+from octobot_commons.constants import CONFIG_ENABLED_OPTION
 from octobot_commons.enums import PriceIndexes
 from octobot_commons.logging.logging_util import get_logger
 from octobot_commons.timestamp_util import is_valid_timestamp
@@ -72,7 +72,7 @@ class ExchangeManager(Initializable):
         self.exchange_type = None
 
         self.client_symbols = []
-        self.client_time_frames = {}
+        self.client_time_frames = []
 
         self.exchange_config = ExchangeConfig(self)
         self.exchange_personal_data = ExchangePersonalData(self)
@@ -151,6 +151,8 @@ class ExchangeManager(Initializable):
     async def _create_simulated_exchange(self):
         self.exchange = ExchangeSimulator(self.config, self.exchange_type, self, self.backtesting_files)
         await self.exchange.initialize()
+        await self._initialize_simulator_time_frames()
+        self.exchange_config.set_config_time_frame()
         self.exchange_config.set_config_traded_pairs()
         await self._create_exchange_channels()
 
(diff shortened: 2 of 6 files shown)


octobot_trading/data/order.py
```diff
@@ -18,6 +18,8 @@
 from asyncio import Lock
 
 import math
+
+from octobot_backtesting.api.backtesting import get_backtesting_current_time
```
Herklos: I don't think this is a good idea to import backtesting module here.

octobot_trading/exchanges/exchange_manager.py
```diff
@@ -253,21 +258,30 @@ def _load_config_symbols_and_time_frames(self):
...
 
     # TIME FRAMES
     def time_frame_exists(self, time_frame, symbol=None):
         if CONFIG_WILDCARD in self.client_time_frames and not self.client_time_frames[CONFIG_WILDCARD]:
             return False
-        elif CONFIG_WILDCARD in self.client_time_frames or symbol is None:
+        elif CONFIG_WILDCARD in self.client_time_frames and symbol is None:
             return time_frame in self.client_time_frames[CONFIG_WILDCARD]
 
+        elif symbol is None:
```
Herklos: Why do you need a specific check when symbol is None ?

octobot_trading/exchanges/exchange_manager.py
```diff
@@ -151,13 +153,16 @@ def reset_exchange_personal_data(self):
...
+        await self._initialize_simulator_time_frames()
+        self.exchange_config.set_config_time_frame()
         self.exchange_config.set_config_traded_pairs()
         await self._create_exchange_channels()
 
     async def _init_simulated_exchange(self):
         try:
             await self.exchange.modify_channels()
             await self.exchange.create_backtesting_exchange_producers()
+            await start_backtesting(self.exchange.backtesting)
```
Herklos: Why this is not made in `ExchangeSimulator` ? Maybe an `ExchangeSimulator.start()` or `.start_backtesting()` ?

octobot_trading/exchanges/exchange_manager.py
```diff
@@ -16,6 +16,8 @@
 import uuid
 from copy import copy
 
+from octobot_backtesting.api.backtesting import start_backtesting
```
Herklos: I don't think the backtesting API should be call in ExchangeManager. I think ExchangeSimulator is the right class to call Backtesting APIs.

octobot_trading/exchanges/data/exchange_config_data.py
```diff
@@ -107,7 +107,7 @@ def _set_config_time_frame(self):
                 self.traded_time_frames.append(time_frame)
         # add shortest timeframe for realtime evaluators
         client_shortest_time_frame = find_min_time_frame(
-            self.exchange_manager.client_time_frames[CONFIG_WILDCARD], MIN_EVAL_TIME_FRAME)
+            next(iter(self.exchange_manager.client_time_frames.values())), MIN_EVAL_TIME_FRAME)
```
Herklos: Why the first `symbol` or `WILDCARD` key should be the shortest time frame ?

octobot_trading/exchanges/exchange_manager.py
```diff
@@ -255,21 +258,30 @@ def _load_config_symbols_and_time_frames(self):
...
     def time_frame_exists(self, time_frame, symbol=None):
         if CONFIG_WILDCARD in self.client_time_frames and not self.client_time_frames[CONFIG_WILDCARD]:
             return False
-        elif CONFIG_WILDCARD in self.client_time_frames or symbol is None:
+        elif CONFIG_WILDCARD in self.client_time_frames and symbol is None:
             return time_frame in self.client_time_frames[CONFIG_WILDCARD]
 
+        elif symbol is None:
+            # check any client_time_frames value: all time frames should be the same
+            return time_frame in next(iter(self.client_time_frames.values()))
```
Herklos: Why do you check only in the first `symbol` or `WILDCARD` key ?

octobot_trading/exchanges/exchange_manager.py
GuillaumeDSM: Because for now symbol is always None given the current usage of this method (only the time_frame argument is set)

octobot_trading/exchanges/data/exchange_config_data.py
GuillaumeDSM: Since we are in an exchange specific situation, any symbol should have the same available time frames. Exceptions being for backtesting with different importers on the same exchange and for some reason different time frames: in this case we don't need to add the shortest timeframe for visualisation because backtesting is fast.

octobot_trading/exchanges/exchange_manager.py
GuillaumeDSM: Same as previous comment except for the visualization part. Here in case of backtesting with multiple datafiles from the same exchange but with different timeframes, we might want to check them all to see if they are the given timeframe. I can add this test.

octobot_trading/exchanges/data/exchange_config_data.py
Herklos: So if you think there is no need of of dictionnary for `client_time_frames` but a list with the `time_frame` is enough we should go for it.

octobot_trading/exchanges/exchange_simulator.py
```diff
@@ -14,19 +14,23 @@
...
+from octobot_commons.time_frame_manager import get_config_time_frame, find_min_time_frame
 from octobot_trading.channels.exchange_channel import get_chan as get_trading_chan
 from octobot_trading.constants import CONFIG_SIMULATOR, CONFIG_DEFAULT_SIMULATOR_FEES, CONFIG_SIMULATOR_FEES, \
     CONFIG_SIMULATOR_FEES_MAKER, CONFIG_SIMULATOR_FEES_TAKER, CONFIG_SIMULATOR_FEES_WITHDRAW
 from octobot_trading.enums import ExchangeConstantsMarketStatusColumns, ExchangeConstantsMarketPropertyColumns, \
     TraderOrderType, FeePropertyColumns
 from octobot_trading.exchanges.abstract_exchange import AbstractExchange
 from octobot_trading.producers.simulator import UNAUTHENTICATED_UPDATER_SIMULATOR_PRODUCERS, \
-    SIMULATOR_PRODUCERS_TO_DATA_TYPE
+    SIMULATOR_PRODUCERS_TO_DATA_TYPE, OHLCVUpdaterSimulator
```
Herklos: I think this is not very clean to import OHLCVUpdaterSimulator here but I don't know how to fix it.

octobot_trading/exchanges/exchange_manager.py
```diff
@@ -249,28 +251,27 @@ def _load_config_symbols_and_time_frames(self):
         client = self.exchange.client
         if client:
             self.client_symbols = client.symbols
-            self.client_time_frames[CONFIG_WILDCARD] = client.timeframes if hasattr(client, "timeframes") else {}
+            self.client_time_frames = client.timeframes if hasattr(client, "timeframes") else []
```
Herklos: What about a backtesting file with different time_frame foreach symbol ?

octobot_trading/exchanges/exchange_simulator.py
GuillaumeDSM: I will change it with an enum, it will also fix imports of these in updater simulators

octobot_trading/exchanges/exchange_manager.py
GuillaumeDSM: I was thinking about it and I think it's not currently possible. If data are from multiple exchanges which can make different timeframes, they will be stored in different files and use different exchange managers. Do you see a case where we can have different available timeframes in the same file ?

Herklos: Not in the same file but in the same backtesting. We wanted to implement multiple data files backtesting. We can imagine, in this case, that we'll have different time frame from the different data files.

GuillaumeDSM: Right, in this case they either come from the same exchange or different ones. If they come from the same exchange, I think we can consider the same available timeframe for each symbol. If they are from different exchange, they will create different exchange managers that can handle different timeframes on their sides.
If you think we can have multiple datafiles on the same exchange and with different timesframes per symbol, then we should re-implement the dict of timeframes per symbol. I think this case can happen (for ex using a mix of different data files with different sources) but is highly improbable since in backtesting it doesn't make much sense to compare multiple symbol results based on different timeframes (and therefore different evaluation styles etc)

Herklos: Yes it's probably the reason why I implemented the dict of timeframes per symbol. I think we can go for a solution that manage only one backtesting file per exchange for now.

PR #90, opened 2020-01-30
[Backtesting] Independant backtesting update

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Trading/pull/89
requires https://github.com/Drakkar-Software/OctoBot-Backtesting/pull/19
requires https://github.com/Drakkar-Software/OctoBot-Channels/pull/20

PR #93, opened 2020-02-06
Update api 

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Trading/pull/90

### DIFF
--- a/octobot_trading/exchanges/data/exchange_symbols_data.py
+++ b/octobot_trading/exchanges/data/exchange_symbols_data.py
@@ -27,9 +27,12 @@ class ExchangeSymbolsData:
         self.config = exchange_manager.config
         self.exchange_symbol_data = {}
 
-    def get_exchange_symbol_data(self, symbol):
+    def get_exchange_symbol_data(self, symbol, allow_creation=True):
         try:
             return self.exchange_symbol_data[symbol]
-        except KeyError:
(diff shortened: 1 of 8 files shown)


octobot_trading/exchanges/data/exchange_symbols_data.py
```diff
@@ -27,9 +27,12 @@ def __init__(self, exchange_manager):
...
             return self.exchange_symbol_data[symbol]
-        except KeyError:
-            self.exchange_symbol_data[symbol] = ExchangeSymbolData(self.exchange_manager, symbol)
-            return self.exchange_symbol_data[symbol]
+        except KeyError as e:
+            if allow_creation:
+                # warning: should only be called in the async loop thread
+                self.exchange_symbol_data[symbol] = ExchangeSymbolData(self.exchange_manager, symbol)
+                return self.exchange_symbol_data[symbol]
+            raise e
```
Herklos: Do you think there is need for a `raise` here ?

GuillaumeDSM: I think it's better to raise because if ```allow_creation``` is false, the caller want the actual existing ExchangeSymbolData and it's better to stop the subsequent processing right now than checking all the time if the returned thing is not None. This is used very often by the web interface and TA evaluators, I think it's better to let them chose to react to this exception rather than forcing a ```if XXX is not None``` all the time.

PR #94, opened 2020-02-08
update trade api and add is_trade_simulated

GuillaumeDSM:


octobot_trading/api/trades.py
```diff
@@ -30,4 +29,8 @@ def get_total_paid_trading_fees(exchange_manager) -> dict:
 
 
 def get_trade_exchange_name(trade) -> str:
-    return get_order_exchange_name(trade.order)
+    return trade.exchange_manager.get_exchange_name()
+
+
+def is_trade_simulated(trade) -> bool:
```
Herklos: Does order have a `simulate` attribute ?

GuillaumeDSM: Nope

Herklos: Ok, do you think a `simulate` attribute have to be added into `Order` and `Trade` ?

GuillaumeDSM: If we want to save trades between bot instances, I think it's better to add it yep

GuillaumeDSM:
done in https://github.com/Drakkar-Software/OctoBot-Trading/pull/95

PR #96, opened 2020-02-08
Add trading modes into exchange manager

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Trading/pull/93

### DIFF
--- a/octobot_trading/exchanges/exchange_manager.py
+++ b/octobot_trading/exchanges/exchange_manager.py
@@ -66,6 +66,7 @@ class ExchangeManager(Initializable):
 
         self.trader = None
         self.exchange = None
+        self.trading_modes = []
 
         self.exchange_web_socket = None
         self.exchange_type = None
@@ -81,6 +82,8 @@ class ExchangeManager(Initializable):
         await self.create_exchanges()
 
     async def stop(self):
+        for trading_mode in self.trading_modes:
+            await trading_mode.stop()
         if self.exchange is not None:
             await self.exchange.stop()
             Exchanges.instance().del_exchange(self.exchange.name, self.id)
@@ -94,6 +97,7 @@ class ExchangeManager(Initializable):
         self.exchange_personal_data = None
         self.exchange_symbols_data = None
         self.trader = None
+        self.trading_modes = []
 
     async def stop_exchange_channels(self):
         try:
--- a/octobot_trading/api/modes.py
+++ b/octobot_trading/api/modes.py
@@ -31,6 +31,18 @@ def init_trading_mode_config(config, trading_tentacles_path) -> None:
     create_classes_list(config, AbstractTradingMode)
 
 
+def get_trading_modes(exchange_manager) -> list:
+    return exchange_manager.trading_modes
+
+
+def get_trading_mode_symbol(trading_mode) -> list:
+    return trading_mode.symbol
+
+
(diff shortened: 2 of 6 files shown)


Herklos:
This is only used for `Interface` purpose ? I know it's not simple to implement this feature but I don't really like to have an attribute in a class of `Trading` only dedicated to `Interface` module...

GuillaumeDSM:
> This is only used for `Interface` purpose ? I know it's not simple to implement this feature but I don't really like to have an attribute in a class of `Trading` only dedicated to `Interface` module...

For now it's only used by interfaces but it can be useful in the future to keep track of the trading modes that are instanciated for an exchange. For now they only keep existing as producers and/or consumers of a channel if they are registered there (which depends on the implementation of the mode), like evaluators. For evaluators it's not really a problem since their only usefull thing: their evaluation is stored in the matrix. For trading modes, we might want to reach them sometimes and there is not easy way to do it.
I think exchange manager is not a bad place to put them since their are created in exchange_builder.

octobot_trading/exchanges/exchange_manager.py
```diff
@@ -94,6 +95,7 @@ def __init__(self, config, exchange_class_string):
         self.exchange_personal_data = None
         self.exchange_symbols_data = None
         self.trader = None
+        self.trading_modes = []
```
Herklos: We can maybe found a usage for this attribute : 
- In `exchange_manager.stop()` : iterate on the list to call trading mode `stop` method (that should stop consumer)

GuillaumeDSM: We could, it would be a cleaner way than stopping them with all the exchange channels but it's not necessary at the moment, do you think we should add it ? Having them available somewhere will also be useful in the future for trading modes that will require interactions with other trading modes (multi symbol stuff / arbitrage)

Herklos: I think it's necessary to stop them when their `exchange_manager` is stopping.

> Having them available somewhere will also be useful in the future for trading modes that will require interactions with other trading modes (multi symbol stuff / arbitrage)

Yes I don't say we shouldn't add this attribute. I am trying to find some usage for it.

GuillaumeDSM: Right, we need to stop them with their ```exchange_manager```, it's already done in ```stop_exchange_channels``` called in ```stop``` but this is only valid if a trading mode is registered as an exchange channel (which should always be the case but is not forced). Therefore I think you are right, we can call a stop for exchange managers there

GuillaumeDSM: I didn't check before but ```AbstractTradingMode``` doesn't have a ```stop``` method nor a consumer/producer list. What do you think about adding them in ```AbstractTentacle``` and fill them with the returned list from ```create_producers``` and ```create_consumers``` ?

Herklos: Why do you want to add a `stop` method in `AbstractTentacle` ? Why not just in `AbstractTradingMode` ?

GuillaumeDSM: Wrong copy-paste, I meant ```AbstractTradingMode```

Herklos: Ok :+1:

PR #98, opened 2020-02-09
Fix typing issues 

GuillaumeDSM:


GuillaumeDSM:
> What about double ?

Tried it, did not work x)

GuillaumeDSM:
> > What about double ?
> 
> Tried it, did not work x)

Now works after migrating everything :)

tests/trades/test_trade_factory.py
```diff
@@ -135,7 +135,7 @@ class TestTradeFactory:
 
         exec_time = time.time()
         trade = create_trade_from_order(order, executed_time=exec_time)
-        assert(exec_time - 50 <= trade.executed_time <= exec_time + 50)
+        assert trade.executed_time == exec_time
```
Herklos: :+1: :+1:

PR #99, opened 2020-02-10
use candle close price to compute recent trades in backtesting

GuillaumeDSM:


octobot_trading/producers/simulator/recent_trade_updater_simulator.py
```diff
@@ -60,9 +60,10 @@ def __init__(self, channel, importer):
     async def _recent_trades_from_ohlcv_callback(self, exchange: str, exchange_id: str,
                                                  symbol: str, time_frame, candle):
         if candle:
-            last_candle_open_price = candle[PriceIndexes.IND_PRICE_OPEN.value]
+            # candles are pushed when completed therefore the current price is the candle's close price
+            last_candle_close_price = candle[PriceIndexes.IND_PRICE_CLOSE.value]
```
Herklos: The last candle is not the current candle but the last completed candle right ?

GuillaumeDSM: Right, ohlcv is pushing candles when they are completed

PR #102, opened 2020-02-16
[ExchangeManager] Register as early as possible

GuillaumeDSM:


octobot_trading/exchanges/exchange_builder.py
```diff
@@ -43,6 +43,9 @@ def __init__(self, config, exchange_name):
     Build
     """
     async def build(self):
+        # first add to global exchanges to always be able to find it later
+        Exchanges.instance().add_exchange(self.exchange_manager, self._matrix_id)
+
```
Herklos: So it'll add an exchange manager not initialized to Exchanges list right?
I' m thinking about the impact of this change.

GuillaumeDSM: Right, the thing is that if we don't do it before initialization, in case of an error in init, we won't be able to find the exchange manager anymore because it hasn't yet been registered

Herklos: Ok but why do we need to have an exchange that fail to initialize in Exchanges ?

GuillaumeDSM: To be able to stop it :)

Herklos: Ok but do we stop it right after this add ? Because I think we shouldn't keep a non-initialized Exchange in Exchanges list.

GuillaumeDSM: We stop it at the start of the next backtesting in case of a backtesting run and after profitability fetching in strategy optimizer, don't keep it. 
The idea here is to be able to call ```stop()``` on the created exchange manager because since we (normally) don't know what happened in case of an exception, it's possible that channels and other stuff got created during initialization, it's important to be able to stop them. Another possibility would be to call stop() directly from ```ExchangeBuilder``` but it would make it much harder to work on in backtesting and optimizer since stop will clean up everything before fetching any result.

Herklos: Ok I see but I really want to avoid having an Exchange that is not properly initialized in Exchanges. As you said, we have no idea if there is channels created or other things linked to this exchange. So I think the best way is to not add an Exchange that is not properly initialize in Exchanges and to `stop()` it directly from ExchangeBuilder to avoid confusion. What do you think about this ?

GuillaumeDSM:
replaced by https://github.com/Drakkar-Software/OctoBot-Trading/pull/108

PR #108, opened 2020-02-17
stop exchange manager on build error 

GuillaumeDSM:


octobot_trading/exchanges/exchange_builder.py
```diff
@@ -43,25 +43,30 @@ def __init__(self, config, exchange_name):
     Build
     """
     async def build(self):
-        await self.exchange_manager.initialize()
+        try:
```
Herklos: To be cleaner, can `async def build(self)` call an `async def _build(self)` as followed ? 
```
async def build(self):
        try:
                await self._build()
        except Exception as e:
            # stop exchange manager if an exception occurred when building it
            await self.exchange_manager.stop()
            raise e        
```
With  `async def _build(self)` that contains the previous  `async def build(self)` code.

octobot_trading/exchanges/exchange_builder.py
```diff
@@ -43,6 +43,15 @@ def __init__(self, config, exchange_name):
     Build
     """
     async def build(self):
+        try:
+            await self._build_exchange_manager()
+        except Exception as e:
+            # stop exchange manager if an exception occurred when building it
+            await self.exchange_manager.stop()
+            raise e
```
Herklos: Should we try to removed if added ?

GuillaumeDSM: It's already removed in ```exchange_manager.stop`()```

PR #118, opened 2020-03-05
[CI] Add pylint 

Herklos:


Herklos:
It also requires https://github.com/Drakkar-Software/OctoBot-Websockets/pull/22

PR #124, opened 2020-03-09
[TradeHistory] Fix parsing when side is None

Herklos:


octobot_trading/orders/order_factory.py
```diff
@@ -30,7 +30,7 @@ def create_order_instance_from_raw(trader, raw_order):
 
 
 def create_order_from_type(trader, order_type):
-    return TraderOrderTypeClasses[order_type](trader)
+    return TraderOrderTypeClasses[order_type](trader) if order_type is not None else Order(trader)
```
GuillaumeDSM: is it really safe to create an ```Order``` ?

Herklos: Yes in some cases it can be unsafe. Do you have an idea for this issue?

GuillaumeDSM: In which case can order_type be None ?

Herklos: In the current case, I was loading trade history from bitmex and some side were empty. So it couldn't figure out which order_type to use to create the Order for the Trade instance creation.

GuillaumeDSM: I see, do you know why the side is empty ?

Herklos: No, I took at look at the corresponding ccxt code. There is nothing problematic there. There is only approx 50% trades concerned by this issue. I didn't had this problem with another exchange.

GuillaumeDSM: Are they getting updated with the right data after some time or it's about historical data ?

Herklos: It's historical data. It's at startup when the bot is collecting historical trades.

PR #129, opened 2020-03-14
[PositionUpdater] Only update one position per symbol 

Herklos:


octobot_trading/producers/positions_updater.py
```diff
@@ -27,26 +27,49 @@ class PositionsUpdater(PositionsProducer):
...
+            self.should_use_open_position_per_symbol = True
+            await self.fetch_position_per_symbol()
+
         while not self.should_stop and not self.channel.is_paused:
+            await asyncio.sleep(self.POSITIONS_REFRESH_TIME)
             try:
-                positions: list = await self.channel.exchange_manager.exchange.get_open_position()
-                if positions:
-                    await self.push(positions, is_closed=False, is_liquidated=False)
+                if self.should_use_open_position_per_symbol:
```
GuillaumeDSM: Why this need for the 2 cases ?

Herklos: To optimize some exchange position update.
Some have an API to get position per symbol and some other has only a fetch all positions API. To prevent exchange spamming and improve position fetch I set an attribute to use "per symbol" way or the full symbol one.

PR #131, opened 2020-03-14
[Funding] Add implementation 

Herklos:
Depends on https://github.com/Drakkar-Software/OctoBot-Trading/pull/130


octobot_trading/producers/funding_updater.py
```diff
@@ -0,0 +1,79 @@
...
+
+class FundingUpdater(FundingProducer):
+    CHANNEL_NAME = FUNDING_CHANNEL
+
+    FUNDING_REFRESH_TIME = 2 * HOURS_TO_SECONDS
+    FUNDING_REFRESH_TIME_MIN = 0.2 * HOURS_TO_SECONDS
+    FUNDING_REFRESH_TIME_MAX = 8 * HOURS_TO_SECONDS
+
+    def __init__(self, channel):
+        super().__init__(channel)
```
GuillaumeDSM: Unnecessary code

octobot_trading/producers/funding_updater.py
```diff
@@ -0,0 +1,79 @@
...
+                                        timestamp=funding[ExchangeConstantsFundingColumns.TIMESTAMP.value])
+            except NotSupported:
+                self.logger.warning(f"{self.channel.exchange_manager.exchange_name} is not supporting updates")
+                await self.pause()
+            except Exception as e:
+                sleep_time = self.FUNDING_REFRESH_TIME_MIN
+                self.logger.exception(e, True, f"Fail to update funding rate : {e}")
+            finally:
+                if next_funding_time:
+                    should_sleep_time = next_funding_time - time.time()
```
GuillaumeDSM: We should add a try/except in the updater loop otherwise if there is an error with the 1st founding update, the other ones will never be updated before the next update found (but they still occured on exchange)

octobot_trading/util/__init__.py
```diff
@@ -78,3 +81,11 @@ def get_market_pair(config, currency) -> (str, bool):
 def get_reference_market(config) -> str:
     # The reference market is the currency unit of the calculated quantity value
     return get_value_or_default(config[CONFIG_TRADING], CONFIG_TRADER_REFERENCE_MARKET, DEFAULT_REFERENCE_MARKET)
+
+
+def datetime_to_timestamp(date_time, date_time_format):
+    return time.mktime(create_datetime(date_time, date_time_format).timetuple())
```
GuillaumeDSM: I think this and ```create_datetime``` could be in https://github.com/Drakkar-Software/OctoBot-Commons/blob/master/octobot_commons/timestamp_util.py. What do you think about it ?

octobot_trading/util/__init__.py
```diff
@@ -78,3 +81,11 @@ def get_market_pair(config, currency) -> (str, bool):
...
+
+
+def create_datetime(date_time, date_time_format):
```
GuillaumeDSM: Not use I understand the purpose of this function: create a ```datetime.datetime``` from the ```date_time``` argument ? I think we should name the argument differently.

octobot_trading/channels/funding.pxd
```diff
@@ -0,0 +1,25 @@
...
+#  You should have received a copy of the GNU Lesser General Public
+#  License along with this library.
+
+from octobot_trading.channels.exchange_channel cimport ExchangeChannel, ExchangeChannelProducer
+
+
+cdef class FundingProducer(ExchangeChannelProducer):
+    pass
+
+cdef class FundingChannel(ExchangeChannel):
```
GuillaumeDSM: 💯

PR #134, opened 2020-03-15
[ExchangeTypes] Add future and spot exchanges 

Herklos:


octobot_trading/exchanges/exchange_manager.py
```diff
@@ -160,9 +163,14 @@ def reset_exchange_personal_data(self):
 
     async def _create_real_exchange(self):
         # create REST based on ccxt exchange
-        if self.is_margin:
+        if self.is_spot_only:
+            await self._search_and_create_spot_exchange()
+        elif self.is_future:
+            await self._search_and_create_future_exchange()
+        elif self.is_spot_only:
```
GuillaumeDSM: if **self.is_spot_only** else if **self.is_spot_only** ?

Herklos: Oh yes it should be` elif self.is_margin` 👍

PR #148, opened 2020-04-05
WIP [CandleManager] Return as array

Herklos:


octobot_trading/data_manager/candles_manager.py
```diff
@@ -172,9 +172,9 @@ def _extract_limited_data(self, data, limit=-1, max_limit=-1):
...
                 return data
-            return data[:max_handled_limit]
+            return np.asarray(data[:max_handled_limit], dtype=np.float64)
 
         if max_limit == -1:
-            return data[-min(limit, len(data)):]
+            return np.asarray(data[-min(limit, len(data)):], dtype=np.float64)
         else:
-            return data[max_handled_limit - limit:max_handled_limit]
+            return np.asarray(data[max_handled_limit - limit:max_handled_limit], dtype=np.float64)
```
GuillaumeDSM, line 180: We should return ```return np.asarray(data[max(0, max_handled_limit - limit):max_handled_limit], dtype=np.float64)``` to handle cases when max_handled_limit is lower than limit

PR #153, opened 2020-04-06
[Channels] Fix time frame filters 

Herklos:


Herklos:
@GuillaumeDSM can you test it please ?

PR #158, opened 2020-04-08
Fix cython headers 

Herklos:
### Fixed
- CandleManager cython headers
- AbstractModeConsumer cython headers

### DIFF
--- a/octobot_trading/data_manager/candles_manager.py
+++ b/octobot_trading/data_manager/candles_manager.py
@@ -168,13 +168,13 @@ class CandlesManager(Initializable):
             self.reached_max = True
 
     def _extract_limited_data(self, data, limit=-1, max_limit=-1):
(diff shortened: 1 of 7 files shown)


octobot_trading/data_manager/candles_manager.py
```diff
@@ -168,13 +168,13 @@ def _inc_candle_index(self):
...
-            return data[:max_handled_limit]
+                return np.array(data)
+            return np.array(data[:max_handled_limit])
 
         if max_limit == -1:
-            return data[-min(limit, len(data)):]
+            return np.array(data[-min(limit, len(data)):])
         else:
-            return data[max_handled_limit - limit:max_handled_limit]
+            return np.array(data[max_handled_limit - limit:max_handled_limit])
```
GuillaumeDSM, line 180: ```return np.array(data[max(0, max_handled_limit - limit):max_handled_limit])``` would be better i think

PR #196, opened 2020-05-12
Multiple fixes

GuillaumeDSM:


octobot_trading/exchanges/exchange_builder.py
```diff
@@ -82,7 +82,6 @@ async def _build_trader(self):
 
     async def _build_modes(self):
         try:
-            init_trading_mode_config(self.config)
```
Herklos, line 85: Managed by tentacle manager ?

octobot_trading/exchanges/exchange_manager.py
```diff
@@ -350,7 +350,7 @@ def _load_config_symbols_and_time_frames(self):
         client = self.exchange.client
         if client:
             self.client_symbols = client.symbols
-            self.client_time_frames = list(client.timeframes.values()) if hasattr(client, "timeframes") else []
+            self.client_time_frames = list(client.timeframes) if hasattr(client, "timeframes") else []
```
Herklos, line 353: We keep only keys now right ?

GuillaumeDSM: yes, list(dict) returns a list of its keys

octobot_trading/exchanges/exchange_builder.py
GuillaumeDSM: By OctoBot, it's related to the advanced manager (when we figure out advanced classes). Maybe we should move it into the tentacle manager if we decide it's its role to manage advanced classes in the future

Herklos: Yes it should be better in Tentacle manager, I don't think Octobot should do that.

GuillaumeDSM: related to https://github.com/Drakkar-Software/OctoBot-Tentacles-Manager/issues/11

PR #200, opened 2020-05-14
[ExchangeChannel] Remove unused _check_producers_state

Herklos:


octobot_trading/channels/exchange_channel.py
```diff
@@ -81,16 +81,6 @@ async def new_consumer(self,
         await self._check_producers_state()
```
GuillaumeDSM, line 81: Default version is doing the job here ?

GuillaumeDSM: actually it's the same code, so yes I guess :)

Herklos: Exactly :wink:

Herklos: It was necessary when the comment was written because it was `__check_producers_state` and now it's `_check_producers_state`.

PR #206, opened 2020-05-16
Real trading fixes

GuillaumeDSM:
can now create and handle filled real orders

### DIFF
--- a/octobot_trading/data_manager/orders_manager.py
+++ b/octobot_trading/data_manager/orders_manager.py
@@ -56,13 +56,14 @@ class OrdersManager(Initializable):
             return True, False
         return self._update_order_from_raw(self.orders[order_id], raw_order), True
 
-    def upsert_order_close(self, order_id, raw_order) -> bool:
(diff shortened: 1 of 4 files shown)


Herklos:
You should update cython header

GuillaumeDSM:
> You should update cython headers

done on the commit right before your comment :)

octobot_trading/data_manager/orders_manager.py
```diff
@@ -56,13 +56,14 @@ def upsert_order(self, order_id, raw_order) -> (bool, bool):
...
+    def upsert_order_close(self, order_id, raw_order) -> (bool, Order):
         if self.has_order(order_id):
+            order = self.orders[order_id]
             self._update_order_from_raw(self.orders[order_id], raw_order)
-            # TODO order -> trade
             self.orders.pop(order_id)
-            return True
-        return False
+            # trade creation should be handled right after this
+            return True, order
```
Herklos, line 65: It can only be `True` if the order existed so I think we can simplify by returning only `order` or `None` and changed will be equals to `order is not None`.

PR #218, opened 2020-05-19
register to channels unique exchange id

GuillaumeDSM:


GuillaumeDSM:
> Do we need to keep the registration to all available exchange ids? I mean to add this new method instead of replacing the old one

It's not used anywhere now, its better to just remove it in my opinion

PR #229, opened 2020-05-25
refactor close and cancel order processes

GuillaumeDSM:


octobot_trading/channels/orders.py
```diff
@@ -43,11 +43,12 @@ async def perform(self, orders, is_closed=False, is_from_bot=True):
                             order,
                             should_notify=False)
                     else:
-                        changed, is_updated = await self.channel.exchange_manager.exchange_personal_data.handle_order_update(
-                            symbol,
-                            order_id,
-                            order,
-                            should_notify=False)
+                        changed, is_updated = \
```
Herklos, line 46: Shouldn't we avoid having 2 return values? What do you think ?

octobot_trading/data/order.py
```diff
@@ -35,7 +34,7 @@ def __init__(self, trader):
         self.trader = trader
         self.exchange_manager = trader.exchange_manager
         self.status = OrderStatus.OPEN
-        self.creation_time = time.time()
+        self.creation_time = self.exchange_manager.exchange.get_exchange_current_time()
```
Herklos, line 37: :+1:

octobot_trading/data/trade.py
```diff
@@ -25,7 +22,7 @@ def __init__(self, trader):
         self.exchange_manager = trader.exchange_manager
 
         self.status = OrderStatus.OPEN
-        self.creation_time = time.time()
+        self.creation_time = self.exchange_manager.exchange.get_exchange_current_time()
```
Herklos, line 25: :+1:

octobot_trading/data_manager/orders_manager.pxd
```diff
@@ -40,8 +40,8 @@ cdef class OrdersManager(Initializable):
 
     cpdef void update_order_attribute(self, str order_id, str key, object value)
     cpdef Order get_order(self, str order_id)
-    cpdef tuple upsert_order(self, str order_id, dict raw_order)
-    cpdef Order upsert_order_close(self, str order_id, dict raw_order)
+    cpdef bint upsert_order_from_raw(self, str order_id, dict raw_order)
```
Herklos, line 43: :+1:

octobot_trading/traders/trader.py
```diff
@@ -145,48 +154,122 @@ async def _create_not_loaded_order(self, new_order: Order, portfolio) -> Order:
 
         return new_order
 
-    async def cancel_order(self, order: Order):
+    async def close_filled_order(self, order):
```
Herklos, line 157: :+1:

octobot_trading/traders/trader.py
```diff
@@ -145,48 +154,122 @@ async def _create_not_loaded_order(self, new_order: Order, portfolio) -> Order:
...
+        # update portfolio with filled order
+        async with self.exchange_manager.exchange_personal_data.get_order_portfolio(order).lock:
+            await self.exchange_manager.exchange_personal_data.handle_portfolio_update_from_order(order)
+
+        # add to trade history and notify
+        await self.exchange_manager.exchange_personal_data.handle_trade_instance_update(
+            create_trade_from_order(order))
+
+        # remove order from open_orders
+        self.exchange_manager.exchange_personal_data.orders_manager.remove_order_instance(order)
```
Herklos, line 177: What do you think of an `remove_order_instance` or `remove_order` in  ExchangePersonnalData ?

octobot_trading/traders/trader.py
```diff
@@ -145,48 +154,122 @@ async def _create_not_loaded_order(self, new_order: Order, portfolio) -> Order:
...
+                if raw_cancelled_order is not None:
+                    await self.exchange_manager.exchange_personal_data.handle_order_update_from_raw(order.symbol,
+                                                                                                    order.order_id,
+                                                                                                    raw_cancelled_order)
+
+    async def _handle_order_cancellation(self, order: Order, is_cancelled_from_exchange: bool) -> dict:
+        raw_cancelled_order = None
+        # if real order: cancel on exchange
+        if not self.simulate and not order.is_self_managed() and not is_cancelled_from_exchange:
+            raw_cancelled_order = await self.exchange_manager.exchange.cancel_order(order.order_id, order.symbol)
```
Herklos, line 210: Should we check something on the exchange return ?

octobot_trading/traders/trader.py
```diff
@@ -145,48 +154,122 @@ async def _create_not_loaded_order(self, new_order: Order, portfolio) -> Order:
...
     async def cancel_open_orders(self, symbol, cancel_loaded_orders=True):
-        # use a copy of the list (not the reference)
-        for order in copy.copy(self.exchange_manager.exchange_personal_data.orders_manager.get_open_orders()):
+        """
+        Should be called only if the goal is to cancel all open orders for a given symbol
+        :param symbol: The symbol to cancel all orders on
+        :param cancel_loaded_orders: When True, also cancels loaded orders (order that are not from this bot instance)
+        :return: None
+        """
+        for order in self.exchange_manager.exchange_personal_data.orders_manager.get_open_orders():
```
Herklos, line 248: Why not iterating on copy of `get_open_orders()` anymore ?

octobot_trading/traders/trader.py
```diff
@@ -145,48 +154,122 @@ async def _create_not_loaded_order(self, new_order: Order, portfolio) -> Order:
...
                 await self.cancel_open_orders(symbol)
 
     async def cancel_all_open_orders(self):
-        # use a copy of the list (not the reference)
-        for order in copy.copy(self.exchange_manager.exchange_personal_data.orders_manager.get_open_orders()):
+        """
+        Cancel all open orders registered on this bot.
+        :return: None
+        """
+        for order in self.exchange_manager.exchange_personal_data.orders_manager.get_open_orders():
```
Herklos, line 270: Same here about the copy

octobot_trading/channels/orders.py
GuillaumeDSM: absolutely, I forgot to update this part, handle_order_update_from_raw is returning only one value

octobot_trading/traders/trader.py
GuillaumeDSM: Not sure we should add it: ExchangePersonnalData only does data update with additional processing (channel publish etc), adding a proxy just for this method would make it confusing regarding what is getting updated I'm affraid

octobot_trading/traders/trader.py
GuillaumeDSM: get_open_orders() already sends a new list, it's not the underlying order dict anymore in v4

octobot_trading/traders/trader.py
GuillaumeDSM: It's the same as when we create an order on exchange, we need to decide where the checks are processed (on whether the command worked or not and ensure the complete order is returned). I think it makes more sense to handle it in rest exchange and maybe raise an exception if there is an issue there ? We need to discuss it I think

octobot_trading/traders/trader.py
Herklos: Right :+1:

octobot_trading/traders/trader.py
Herklos: Yes we will need something like this in the rest exchange. You're right we should add this check with the one concerning the order creation in an other PR.

octobot_trading/traders/trader.py
Herklos: Oh yes you're right :+1:

PR #228, opened 2020-05-24
Add price events manager

Herklos:
Done in PR : 
- Add price events manager implementation
- Add recent trades manager tests
- Add price events manager tests

PR #231, opened 2020-05-26
update market status fixer

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Trading/pull/228 for CYTHON_IGNORE

### DIFF
--- a/octobot_trading/exchanges/util/exchange_market_status_fixer.pxd
+++ b/octobot_trading/exchanges/util/exchange_market_status_fixer.pxd
@@ -16,25 +16,24 @@
 #  License along with this library.
 
 cpdef bint is_ms_valid(object value, bint zero_valid=*)
-
-cdef void calculate_costs(dict market_limit)
-cdef void calculate_prices(dict market_limit)
-cdef void calculate_amounts(dict market_limit)
-cdef object get_markets_limit(dict market_limit)
-cdef void fix_market_status_limits_from_current_data(dict market_limit)
 cdef bint check_market_status_limits(market_limit)
 cdef bint check_market_status_values(market_limit, bint zero_valid=*)
+cdef object get_markets_limit(dict market_limit)
+cdef void calculate_amounts(dict market_limit)
+cdef void calculate_costs(dict market_limit)
+cdef void calculate_prices(dict market_limit)
+cdef void fix_market_status_limits_from_current_data(dict market_limit)
 
 cdef class ExchangeMarketStatusFixer:
-    cdef public object market_status
-    cdef public object price_example
-    cdef public object market_status_specific
+    cdef public dict market_status
+    cdef object price_example
+    cdef dict market_status_specific
 
-    cdef void __fix_market_status_precision(self)
-    cdef void __fix_market_status_limits(self)
-    cdef object __calculate_amount(self)
-    cdef void __fix_market_status_limits_with_price(self)
-    cdef double __get_price_precision(self)
-    cdef void __fix_market_status_precision_with_price(self)
-    cdef void __fix_market_status_precision_with_specific(self)
-    cdef void __fix_market_status_limits_with_specific(self)
+    cdef void _fix_market_status_precision(self)
+    cdef void _fix_market_status_limits(self)
+    cdef tuple _calculate_amount(self)
+    cdef void _fix_market_status_limits_with_price(self)
+    cdef void _fix_market_status_precision_with_price(self)
+    cdef double _get_price_precision(self)
+    cdef void _fix_market_status_limits_with_specific(self)
+    cdef void _fix_market_status_precision_with_specific(self)
--- a/octobot_trading/exchanges/util/exchange_market_status_fixer.py
+++ b/octobot_trading/exchanges/util/exchange_market_status_fixer.py
@@ -135,10 +135,10 @@ class ExchangeMarketStatusFixer:
         else:
             self.market_status_specific = None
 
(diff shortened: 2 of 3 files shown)


octobot_trading/exchanges/util/exchange_market_status_fixer.pxd
```diff
@@ -16,25 +16,26 @@
...
-    cdef object __calculate_amount(self)
-    cdef void __fix_market_status_limits_with_price(self)
-    cdef double __get_price_precision(self)
-    cdef void __fix_market_status_precision_with_price(self)
-    cdef void __fix_market_status_precision_with_specific(self)
-    cdef void __fix_market_status_limits_with_specific(self)
+    cpdef double _get_price_precision(self)
+    cpdef void _fix_market_status_precision_with_price(self)
+    cpdef void _fix_market_status_limits_with_specific(self)
+    cpdef void _fix_market_status_limits_with_price(self)
```
GuillaumeDSM, line 36: I had to make these cpdef to be able to call them in tests, what do you think about it @Herklos, should we keep cdef and avoid testing them directly ?

Herklos: We shouldn't change the cython header for test purposes. For similar cases I used to use CYTHON_IGNORE condition to pass the test when we are testing in compiled way.

octobot_trading/exchanges/util/exchange_market_status_fixer.pxd
```diff
@@ -16,25 +16,26 @@
...
 cdef void calculate_costs(dict market_limit)
 cdef void calculate_prices(dict market_limit)
 cdef void calculate_amounts(dict market_limit)
 cdef object get_markets_limit(dict market_limit)
 cdef void fix_market_status_limits_from_current_data(dict market_limit)
-cdef bint check_market_status_limits(market_limit)
-cdef bint check_market_status_values(market_limit, bint zero_valid=*)
 
 cdef class ExchangeMarketStatusFixer:
     cdef public object market_status
```
Herklos, line 29: Should it be a dict?

octobot_trading/exchanges/util/exchange_market_status_fixer.pxd
```diff
@@ -16,25 +16,26 @@
...
     cdef public object price_example
```
Herklos, line 30: I think it's a double

Herklos: And it shouldn't be public

octobot_trading/exchanges/util/exchange_market_status_fixer.pxd
```diff
@@ -16,25 +16,26 @@
...
     cdef public object market_status_specific
```
Herklos, line 31: I think it's also a dict. I don't know if we are using it publicly

octobot_trading/exchanges/util/exchange_market_status_fixer.pxd
Herklos: > I think it's a double

Don't forget the double :smile:

octobot_trading/exchanges/util/exchange_market_status_fixer.pxd
Herklos: > We shouldn't change the cython header for test purposes. For similar cases I used to use CYTHON_IGNORE condition to pass the test when we are testing in compiled way.

To use CYTHON_IGNORE you need https://github.com/Drakkar-Software/OctoBot-Trading/pull/228

octobot_trading/exchanges/util/exchange_market_status_fixer.pxd
```diff
@@ -16,25 +16,24 @@
...
 cpdef bint is_ms_valid(object value, bint zero_valid=*)
-
-cdef void calculate_costs(dict market_limit)
-cdef void calculate_prices(dict market_limit)
-cdef void calculate_amounts(dict market_limit)
-cdef object get_markets_limit(dict market_limit)
-cdef void fix_market_status_limits_from_current_data(dict market_limit)
-cdef bint check_market_status_limits(market_limit)
-cdef bint check_market_status_values(market_limit, bint zero_valid=*)
+cpdef bint check_market_status_limits(market_limit)
```
Herklos, line 19: Why do we need this functions to be public ?

GuillaumeDSM: every commit after tmp is a test, I'm trying to figure out why the compiled version is segfaulting

GuillaumeDSM: it looks like when making these private, python is enable to import them from the ExchangeMarketStatusFixer class

Herklos: :open_mouth:

GuillaumeDSM: further tests in progess

GuillaumeDSM: Issue fixed, a function was imported in another module

octobot_trading/exchanges/util/exchange_market_status_fixer.pxd
GuillaumeDSM: With it being initialized bu default at None, it has to be object

PR #241, opened 2020-06-06
[Coverage] Add cli and api folders ignore

Herklos:


.coveragerc
```diff
@@ -1,7 +1,10 @@
 [run]
 omit =
+    octobot_trading/cli/*
+    octobot_trading/api/*
     tests/*
     tentacles/*
     logs/*
     backtesting/*
     demo.py
+    setup.py
```
GuillaumeDSM, line 10: why setup.py ?

Herklos: Do we want to test setup.py ?

GuillaumeDSM: Right 😂

PR #243, opened 2020-06-06
Handle non trading exchanges

GuillaumeDSM:


octobot_trading/data/order.py
```diff
@@ -253,11 +254,11 @@ def get_profitability(self):
         if self.filled_price != 0 and self.created_last_price != 0:
             if self.filled_price >= self.created_last_price:
                 self.order_profitability = 1 - self.filled_price / self.created_last_price
-                if self.side == TradeOrderSide.SELL:
+                if self.side is TradeOrderSide.SELL:
```
Herklos, line 257: :+1:

octobot_trading/exchanges/exchange_builder.py
```diff
@@ -63,7 +69,10 @@ async def _build_exchange_manager(self):
 
             # create trading modes
             if self._is_using_trading_modes:
-                self.exchange_manager.trading_modes = await self._build_modes()
+                if self.exchange_manager.is_trading:
+                    self.exchange_manager.trading_modes = await self._build_trading_modes(trading_mode_class)
+                else:
+                    self.logger.info(f"{self.exchange_name} exchange is online and won't be trading")
```
Herklos, line 75: What do you think about "exchange is online **but** won't be trading"

Herklos:
It would have been better to split into two PR, one for the commit on orders and the other one for non trading exchanges

GuillaumeDSM:
> It would have been better to split into two PR, one for the commit on orders and the other one for non trading exchanges

Yep, I will rework it when I'm done with the arbitrage trading mode

octobot_trading/traders/trader.py
```diff
@@ -51,13 +51,13 @@ def __init__(self, config, exchange_manager):
 
         if not hasattr(self, 'simulate'):
             self.simulate = False
-
         self.is_enabled = self.__class__.enabled(self.config)
-        self.logger.debug(f"{'Enabled' if self.is_enabled else 'Disabled'} on {self.exchange_manager.exchange_name}")
 
     async def initialize_impl(self):
+        self.is_enabled = self.is_enabled and self.exchange_manager.is_trading
```
Herklos, line 57: Finally we disable the trader if the exchange manager is not trading ?

octobot_trading/exchanges/exchange_builder.py
GuillaumeDSM: I was also thinking about this option and at the end I think **and** is better because it's a normal behaviour, it's not like if there is an issue somewhere. Usually adding **but** means something is negative, here it's not the case.

octobot_trading/traders/trader.py
GuillaumeDSM: I think it makes more sense to have the trader being deactivated when its exchange is not trading rather than having an activated trader on a no-trading exchange

Herklos: Yes it seems logic, I was just thinking about unexpected effects of disabling Trader. I think it should be ok :+1:

GuillaumeDSM: For now I see no issue :)

PR #247, opened 2020-06-08
fix orders on_fill and tests

GuillaumeDSM:


CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.7.3] - 2020-06-09
+### Fixed
+- Stop orders 
+- Tests
+
```
Herklos, line 11: I would prefer to have 1.8.0 with price events manager and new orders. So you current version should be 1.8.1

PR #249, opened 2020-06-10
fix portfolio default value

GuillaumeDSM:
None default value creates exceptions on real trading without being authenticated

### DIFF
--- a/octobot_trading/data_manager/portfolio_manager.pxd
+++ b/octobot_trading/data_manager/portfolio_manager.pxd
@@ -38,6 +38,5 @@ cdef class PortfolioManager(Initializable):
     cpdef bint handle_balance_update_from_order(self, Order order)
     cpdef bint handle_balance_update(self, dict balance)
 
-    cdef void _reset_portfolio(self)
     cdef void _load_portfolio(self)
     cdef void _set_starting_simulated_portfolio(self)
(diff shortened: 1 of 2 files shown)


octobot_trading/data/portfolio.py
```diff
@@ -40,7 +40,7 @@ def __init__(self, exchange_name, is_simulated=False):
 
         self.logger = get_logger(f"{self.__class__.__name__}{'Simulator' if is_simulated else ''}[{exchange_name}]")
         self.lock = Lock()
-        self.portfolio = None
+        self.portfolio = {}
```
Herklos, line 43: ?

Herklos: A portfolio should be initialize and will be set to `{}` during initialize()

GuillaumeDSM: Using None is producing a lot of exceptions at different places when portfolio is never initialized (when exchange creds are wrong / missing for ex)

GuillaumeDSM: I don't thing having a None default value is adding any value to this

Herklos: It doesn't make sense to create a portfolio and not initialize it.

PR #250, opened 2020-06-11
add ohlcv subscribe and is_trading api

GuillaumeDSM:


octobot_trading/api/symbol_data.py
```diff
@@ -87,3 +89,9 @@ def force_set_mark_price(exchange_manager, symbol, price):
 def is_mark_price_initialized(exchange_manager, symbol: str) -> bool:
     return exchange_manager.exchange_symbols_data.get_exchange_symbol_data(symbol).prices_manager.\
         valid_price_received_event.is_set()
+
+
+async def subscribe_to_ohlcv_channel(callback, exchange_id):
```
Herklos, line 94: I don't understand why it's in symbol_data api

octobot_trading/api/exchange.py
```diff
@@ -67,6 +67,10 @@ def get_trading_exchanges(exchange_managers) -> list:
             if exchange_manager.is_trading]
 
 
+def is_exchange_trading(exchange_manager) -> bool:
```
Herklos, line 70: :+1:

octobot_trading/api/symbol_data.py
GuillaumeDSM: It's data related to a traded symbol, where do you see it otherwise ?

GuillaumeDSM: I'm adding a channels.py api

PR #251, opened 2020-06-13
[Order] Fix artificial order creation portfolio usage

Herklos:
Update portfolio from order before creating artificial order

### DIFF
--- a/octobot_trading/orders/types/limit/stop_loss_limit_order.py
+++ b/octobot_trading/orders/types/limit/stop_loss_limit_order.py
@@ -25,14 +25,14 @@ class StopLossLimitOrder(LimitOrder):
         self.trigger_above = False
         self.limit_price = self.UNINITIALIZED_LIMIT_PRICE
 
-    async def on_fill(self):
-        await self.trader.create_artificial_order(TraderOrderType.SELL_MARKET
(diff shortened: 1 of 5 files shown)


octobot_trading/orders/types/limit/stop_loss_limit_order.py
```diff
@@ -25,14 +25,14 @@ def __init__(self, trader, side=TradeOrderSide.SELL):
...
-                                                  if self.side is TradeOrderSide.SELL
-                                                  else TraderOrderType.BUY_MARKET,
-                                                  self.symbol, self.origin_stop_price,
-                                                  self.origin_quantity,
-                                                  self.limit_price
-                                                  if self.limit_price != self.UNINITIALIZED_LIMIT_PRICE else
-                                                  self.origin_stop_price,
-                                                  self.linked_portfolio)
-        await super().on_fill()
+    async def on_trade_creation(self):
```
GuillaumeDSM, line 28: this is to prevent secondary order creation on a cleared order ?

Herklos: Exactly it's called just after portfolio update from the previous Order

PR #242, opened 2020-06-06
[Order] Add trailing stop implementation

Herklos:

PR #254, opened 2020-06-13
[Order] Add trailing stop from get_order_type

Herklos:
Depends on https://github.com/Drakkar-Software/OctoBot-Trading/pull/242


Herklos:
Integrated in #242

PR #248, opened 2020-06-10
Fix exchange init

GuillaumeDSM:
`__str__()` results:
after init
image

during init
image

For later when margin / future wil be available: add these options in exchange config web interface (maybe radio button)

PR #256, opened 2020-06-14
[Order] Synchronize with exchange after creation

Herklos:
Depends on https://github.com/Drakkar-Software/OctoBot-Trading/pull/249
Depends on https://github.com/Drakkar-Software/OctoBot-Trading/pull/248

### DIFF
--- a/octobot_trading/data/order.py
+++ b/octobot_trading/data/order.py
@@ -175,6 +175,12 @@ class Order(Initializable):
         Initialize order status update tasks
         """
         if not self.exchange_manager.is_backtesting:
+            if not self.exchange_manager.is_simulated:
+                # Create a task to synchronize order data with exchange
+                asyncio.create_task(get_chan(ORDERS_CHANNEL, self.exchange_manager.id).get_internal_producer().
+                                    update_order_from_exchange(self))
+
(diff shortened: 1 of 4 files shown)


octobot_trading/data/order.py
```diff
@@ -175,6 +175,12 @@ async def initialize_impl(self):
         Initialize order status update tasks
         """
         if not self.exchange_manager.is_backtesting:
+            if not self.exchange_manager.is_simulated:
+                # Create a task to synchronize order data with exchange
+                asyncio.create_task(get_chan(ORDERS_CHANNEL, self.exchange_manager.id).get_internal_producer().
+                                    update_order_from_exchange(self))
```
GuillaumeDSM, line 181: Not awaiting this task will make it so that the following call (`asyncio.get_event_loop().call_later(self.CHECK_ORDER_STATUS_AFTER_INIT_DELAY...`) will be executed after a delay that is independent from whether this `update_order_from_exchange` task is completed or not, is it what we want in real trading here?

Herklos: There is no link between those two methods, I think it's a correct behavior.

PR #273, opened 2020-06-22
fix candles init index error and infinite exchange spam

GuillaumeDSM:


octobot_trading/producers/ohlcv_updater.py
```diff
@@ -82,15 +85,26 @@ async def _initialize_candles(self, time_frame, pair):
         candles: list = await self.channel.exchange_manager.exchange \
             .get_symbol_prices(pair, time_frame, limit=self.OHLCV_OLD_LIMIT)
         self.channel.exchange_manager.uniformize_candles_if_necessary(candles)
-        if candles:
+        if len(candles) > 1:
```
Herklos, line 88: https://docs.python.org/3/library/stdtypes.html#truth-value-testing

octobot_trading/producers/ohlcv_updater.py
```diff
@@ -82,15 +85,26 @@ async def _initialize_candles(self, time_frame, pair):
...
         else:
             # When candle history cannot be loaded, retry to load it later
-            self.logger.error(f"Failed to init candle history")
-            asyncio.get_event_loop().call_later(self.OHLCV_INITIALIZATION_RETRY_DELAY,
-                                                asyncio.create_task,
-                                                self._initialize_candles(time_frame, pair))
+            self.logger.warning(f"Failed to initialize candle history for {pair} on {time_frame}. Retrying in "
+                                f"{self.OHLCV_INITIALIZATION_RETRY_DELAY} seconds")
+            try:
+                self.candles_init_retry[pair][time_frame] += 1
```
Herklos, line 97: I think we don't care of symbol and tf, if it doesn't work for 1, it wont work for all. We should just limit global retry count in my opinion

octobot_trading/producers/ohlcv_updater.py
GuillaumeDSM: I don"t understand what you mean, here we want to check if the list has a size >= 2

octobot_trading/producers/ohlcv_updater.py
GuillaumeDSM: We can do that but in this case it's more complicated to manage since there is a request for each pair * time_frames, you think we should do that instead ?

octobot_trading/producers/ohlcv_updater.py
Herklos: oh yeah my bad

octobot_trading/producers/ohlcv_updater.py
Herklos: Do you think it'll be more complicated in this way ? I thought it would be less

octobot_trading/producers/ohlcv_updater.py
```diff
@@ -64,72 +65,94 @@ def _get_time_frames(self):
...
             ])
         except Exception as e:
             self.logger.exception(e, True, f"Error while initializing candles: {e}")
         finally:
-            self.logger.debug("Candle history loaded")
+            self.logger.debug("Candle history initial fetch completed")
             self.is_initialized = True
 
-    async def _initialize_candles(self, time_frame, pair):
+    async def _initialize_candles(self, time_frame, pair, retry):
```
Herklos, line 78: I would prefer `should_retry` :wink:

octobot_trading/producers/ohlcv_updater.py
```diff
@@ -64,72 +65,94 @@ def _get_time_frames(self):
...
-                    await asyncio.sleep(should_sleep_time)
+                        await asyncio.sleep(should_sleep_time)
+                    else:
+                        # TODO think about asyncio.call_at or call_later
+                        await asyncio.sleep(time_frame_sleep)
                 else:
-                    # TODO think about asyncio.call_at or call_later
-                    await asyncio.sleep(time_frame_sleep)
+                    # candles on this time frame have not been initialized: sleep until the next candle update
+                    await asyncio.sleep(max(0.0, time_frame_sleep - (time.time() - start_update_time)))
```
Herklos, line 155: :+1:

PR #281, opened 2020-06-28
add requirements for binance websocket

GuillaumeDSM:


octobot_trading/enums.py
```diff
@@ -307,6 +307,7 @@ class WebsocketFeeds(Enum):
     PORTFOLIO = 'portfolio'
     POSITION = 'position'
     TRADE = 'trade'
+    USER_DATA = 'user_data'
```
Herklos, line 310: What is user data? There is no channel for this

octobot_trading/exchanges/types/websocket_exchange.py
```diff
@@ -96,14 +96,15 @@ def _initialize(self, pairs, channels):
...
-    async def _watch(self):
-        if self.last_msg:
-            if datetime.utcnow() - datetime.timedelta(seconds=self.timeout) > self.last_msg:
-                self.logger.warning("No messages received within timeout, restarting connection")
-                await self.reconnect()
-        await self.ping()
-        self.logger.debug("Sending keepalive...")
-        await asyncio.sleep(self.timeout_interval)
+    async def _watcher(self):
+        while True:
```
Herklos, line 100: :+1:

octobot_trading/exchanges/types/websocket_exchange.py
```diff
@@ -309,9 +309,21 @@ def get_exchange_pair(self, pair: str) -> str:
...
+    def safe_integer(self, dictionary, key, default_value=None):
+        return ccxtExchange.safe_integer(dictionary, key, default_value)
+
+    def safe_string(self, dictionary, key, default_value=None, lower=False):
+        value = ccxtExchange.safe_string(dictionary, key, default_value)
+        if value is not None and lower:
+            value = value.lower()
+        return value
+
+    def iso8601(self, timestamp):
```
Herklos, line 324: I don't think it's a good idea to add those methods. Prefer using rest exchange parsers

octobot_trading/enums.py
GuillaumeDSM: Maybe it's not the right place to put it, it's the stream containing all user-related feeds (balance, orders, etc) in binance ws https://github.com/binance-exchange/binance-official-api-docs/blob/master/user-data-stream.md

octobot_trading/exchanges/types/websocket_exchange.py
GuillaumeDSM: What do you mean ? When we can't parse elements from ccxt directly

octobot_trading/enums.py
Herklos: To match the way `WebsocketFeeds` is used in other WS. Prefer using : 
`BALANCE = 'user_data'`
`ORDERS = 'user_data'`

octobot_trading/exchanges/types/websocket_exchange.py
Herklos: You can directly use the exchange client from the WS. I don't really understand why we need a wrapper here

octobot_trading/exchanges/types/websocket_exchange.py
```diff
@@ -119,24 +118,23 @@ async def _connect(self):
...
                     self.websocket = websocket
                     self.on_open()
-                    self._watch_task = asyncio.create_task(self._watch())
-                    # connection was successful, reset retry count and delay
+                    self._watch_task = asyncio.create_task(self._watcher())
+                    # connection was successful, reset delay
                     delay = 1
-                    if not self.exchange_manager.without_auth and self.api_key and self.api_secret:
+                    if not self.exchange_manager.without_auth \
+                            and not self.exchange_manager.is_trader_simulated \
```
Herklos, line 125: Why can we have a WS when simulating trader?

octobot_trading/channels/order_book.py
```diff
@@ -21,14 +21,15 @@
 
 
 class OrderBookProducer(ExchangeChannelProducer):
-    async def push(self, symbol, asks, bids):
-        await self.perform(symbol, asks, bids)
+    async def push(self, symbol, asks, bids, update_order_book=True):
```
Herklos, line 24: :+1:

PR #288, opened 2020-07-05
[OrderBookManager] Add pair [price, size] conversion to order format

Herklos:


tests/util/random_numbers.py
```diff
@@ -53,8 +53,8 @@ def random_funding_rate(min_value=0):
 
 
 def random_order_book_side(min_value=0, count=2):
-    return zip(random_prices(min_value=min_value, count=count),
-               random_quantities(min_value=min_value, count=count))
+    return [list(pair) for pair in zip(random_prices(min_value=min_value, count=count),
```
GuillaumeDSM, line 56: In tests, are `list(...)` calls still required ? Like in:
```python
asks = list(random_order_book_side(count=100))
bids = list(random_order_book_side(count=100))
```

Herklos: Exact :+1:

PR #293, opened 2020-07-19
Force refresh portfolio

GuillaumeDSM:


octobot_trading/channels/orders.py
```diff
@@ -59,6 +62,10 @@ async def perform(self, orders, is_closed=False, is_from_bot=True):
 
             if symbol is not None:
                 await self._check_missing_open_orders(symbol, orders)
+            if possible_new_order:
+                # possibility new order: refresh portfolio to ensure available funds are up to date
+                await self.channel.exchange_manager.exchange_personal_data.portfolio_manager.\
```
Herklos, line 67: There is already a new order detection mechanism. You should put this code in it

GuillaumeDSM: I did at first. The issue with it is that it will trigger a portfolio refresh for each new order, which is not necessary. Here it will only refresh the portfolio once.

Herklos: I think it should be better to refresh at the end of the open orders refresh mechanism if there is a new order.

GuillaumeDSM: In Order Updater ? In this case we need somehow to call it on websocket update as well, which doesn't use open order updater. I put it here because it's the common code where a new order will always end up. However maybe it would be better in open orders update right. And then it would be on the websocket to request a balance update if it's not handling it by its websocket updaters. What do you think about it ?

octobot_trading/data_manager/portfolio_manager.py
```diff
@@ -64,10 +64,18 @@ async def handle_balance_update_from_order(self, order):
                 return True
             else:
                 # on real trading: reload portfolio to ensure portfolio sync
-                return await get_chan(BALANCE_CHANNEL, self.exchange_manager.id).get_internal_producer().\
-                    update_portfolio_from_exchange(self)
+                return await self.refresh_real_trader_portfolio()
         return False
 
+    async def refresh_real_trader_portfolio(self, force_manual_refresh=False) -> bool:
```
Herklos, line 70: not in balance producer ?

PR #304, opened 2020-08-20
fix order states tests coroutine warnings

GuillaumeDSM:
fixes `RuntimeWarning: coroutine 'LimitOrder.update_order_status' was never awaited` warning


Herklos (approved):
:+1:

PR #314, opened 2020-08-27
TEMPORARY Order update comments

GuillaumeDSM:
Here are my thoughts on what to update to ensure orders can only get updated once at a time.
This probably does not contain all required update but can be used to cross check and find forgotten stuff


octobot_trading/producers/orders_updater.py
```diff
@@ -21,6 +21,26 @@
 from octobot_trading.channels.orders import OrdersProducer
 from octobot_trading.constants import ORDERS_CHANNEL
 
+# OpenOrdersUpdater and CloseOrdersUpdater are not triggering automatically like before: they are now triggered via a
```
Herklos, line 24: :+1:

PR #321, opened 2020-09-06
[Trades][Positions] Drop None updates

Herklos:


octobot_trading/channels/positions.py
```diff
@@ -27,6 +27,8 @@ async def push(self, positions, is_closed=False, is_liquidated=False, is_from_bo
     async def perform(self, positions, is_closed=False, is_liquidated=False, is_from_bot=True):
         try:
             for position in positions:
+                if not position:
```
GuillaumeDSM, line 30: a position can be None ?

octobot_trading/channels/trades.py
```diff
@@ -27,6 +27,8 @@ async def push(self, trades, old_trade=False):
     async def perform(self, trades, old_trade=False):
         try:
             for trade in trades:
+                if not trade:
```
GuillaumeDSM, line 30: a trade can be None ?

octobot_trading/channels/positions.py
Herklos: Yes, on binance future, when there is open orders but no position, binance is giving `[None]` instead of `[]` when requesting open positions.

octobot_trading/channels/trades.py
Herklos: Yes bitmex can response a trade history containing None like : `[TradeA, TradeB, None, TradeC]`. I don't know if it's an issue with ccxt but it happens

GuillaumeDSM (approved):

PR #359, opened 2020-09-27
add internal_callback default implementation

GuillaumeDSM:


octobot_trading/consumers/abstract_mode_consumer.py
```diff
@@ -31,8 +31,13 @@ def flush(self):
         self.trading_mode = None
         self.exchange_manager = None
 
-    async def internal_callback(self, **kwargs):
-        raise NotImplementedError("internal_callback is not implemented")
+    async def internal_callback(self, trading_mode_name, cryptocurrency, symbol, time_frame, final_note, state, data):
```
Herklos, line 34: I don't understand why AbstractModeConsumer should implement this method

GuillaumeDSM: For now it's the exact same implementation everywhere in any subclass (in 4 trading modes), I agree that this implementation should not be forced on every trading mode (overriding the method is still possible) however since we now realize that all existing trading mode use this implementation, I think it makes sense to factorize it somehow with a default behavior.

Herklos: OK :+1:

Herklos (approved):

PR #421, opened 2020-11-19
[Exchanges] Refactor implementations

Herklos:
- Add connectors instead of double inheritance between implementations and exchange types


octobot_trading/exchanges/implementations/spot_ccxt_exchange.py
```diff
@@ -43,26 +62,26 @@ async def create_order(self, order_type: enums.TraderOrderType, symbol: str, qua
...
         except ccxt.InsufficientFunds as e:
-            self._log_error(e, order_type, symbol, quantity, price, stop_price)
+            self.connector.log_error(e, order_type, symbol, quantity, price, stop_price)
             self.logger.warning(str(e))
             raise errors.MissingFunds(e)
         except ccxt.NotSupported:
             raise errors.NotSupported
         except Exception as e:
-            self._log_error(e, order_type, symbol, quantity, price, stop_price)
+            self.connector.log_error(e, order_type, symbol, quantity, price, stop_price)
```
GuillaumeDSM, line 71: why logging from connector ?

tests/exchanges/__init__.py
```diff
@@ -50,15 +50,19 @@ async def exchange_manager(request):
...
-    # let updaters gracefully shutdown
-    await wait_asyncio_next_cycle()
+    try:
+        cancel_ccxt_throttle_task()
+        await exchange_manager_instance.stop()
+        # let updaters gracefully shutdown
+        await wait_asyncio_next_cycle()
+    except RuntimeError as e:
+        # prevent to fail test when event loop is closed
+        pass
```
GuillaumeDSM, line 65: :o do we really have to add this ?

GuillaumeDSM (commented):
great work !

octobot_trading/exchanges/implementations/spot_ccxt_exchange.py
Herklos: I wanted to keep the previous behaviour, do you think SpotCCXTExchange should use its own logger ?

GuillaumeDSM: I think it will make it confusing if a code is logging as if it was from somewhere else

Herklos: Yes :+1: I'll add a new logger

GuillaumeDSM (approved):
👍

PR #423, opened 2020-11-21
move is_authenticated to connectors and bind it in exchange manager

GuillaumeDSM:


octobot_trading/exchanges/abstract_exchange.py
```diff
@@ -78,6 +75,9 @@ def is_simulated_exchange(cls) -> bool:
         """
         return False
 
+    def authenticated(self) -> bool:
+        return self.connector.is_authenticated
```
Herklos, line 79: :+1:

Herklos (approved):

PR #426, opened 2020-11-25
Fix order creation checks

GuillaumeDSM:


octobot_trading/exchanges/implementations/spot_ccxt_exchange.py
```diff
@@ -96,12 +100,14 @@ async def _create_specific_order(self, order_type, symbol, quantity, price=None)
             created_order = None
         return created_order
 
-    @staticmethod
-    def _ensure_order_details_completeness(order, order_required_fields=None):
+    def _ensure_order_details_completeness(self, order, order_required_fields=None, order_non_empty_fields=None):
```
Herklos, line 103: Why not : 
```suggestion
    def _ensure_order_details_completeness(self, order, order_required_fields=ORDER_REQUIRED_FIELDS, order_non_empty_fields=ORDER_NON_EMPTY_FIELDS):
``` ?

GuillaumeDSM: Because these are mutable elements that should not be used as default values

Herklos: Oh yes my bad :+1:

Herklos (approved):

PR #436, opened 2020-12-01
fix order creation error on exception

GuillaumeDSM:


octobot_trading/personal_data/orders/order_util.py
```diff
@@ -133,3 +133,9 @@ def parse_order_status(raw_order):
 
 def parse_is_cancelled(raw_order):
     return parse_order_status(raw_order) in {enums.OrderStatus.CANCELED, enums.OrderStatus.CLOSED}
+
+
+def log_order_creation_error(logger, error, order_type, symbol, quantity, price, stop_price):
```
Herklos, line 138: I don't think it's the proper place to put this method. The order creation error is linked to an exchange action, not an OctoBot internal error. If you don't want to keep it in `ccxt_exchange` that can be understood I think it should be put into an exchange util module or a parent of `ccxt_exchange`.

octobot_trading/exchanges/implementations/spot_ccxt_exchange.py
```diff
@@ -23,6 +23,7 @@
 import octobot_trading.errors as errors
 import octobot_trading.exchanges.types as exchanges_types
 import octobot_trading.exchanges.connectors as exchange_connectors
+import octobot_trading.personal_data as personal_data
```
Herklos, line 26: Still needed ?

tests/exchanges/test_abstract_exchange.py
```diff
@@ -0,0 +1,50 @@
...
+EXCHANGE_NAME = "binance"
+
+
+@pytest.fixture
+def abstract_exchange():
+    config = test_config.load_test_config()
+    return exchanges.AbstractExchange(config, exchanges.ExchangeManager(config, EXCHANGE_NAME))
+
+
+async def test_log_order_creation_error(abstract_exchange):
```
Herklos, line 36: :+1:

Herklos (approved):

PR #443, opened 2020-12-20
properly check exchange authentication before starting updaters

GuillaumeDSM:


octobot_trading/exchanges/exchange_manager.py
```diff
@@ -205,8 +205,8 @@ def get_exchange_credentials(self, logger, exchange_name):
 
     @staticmethod
     def handle_token_error(error, logger):
```
Herklos, line 207: I think we should move this method in ccxt_exchange or abstract_exchange

GuillaumeDSM: I move it to abstract_exchange

Herklos (approved):

PR #447, opened 2021-01-04
prevent traded pairs redundancy in exchange config data

GuillaumeDSM:


Herklos (commented):
Can we add a test to make sure it never happens again ?

Herklos (approved):
:+1:

Herklos (approved):

PR #451, opened 2021-01-26
[Version] v1.11.25

Herklos:
A bug is currently impacting all real get_balances. Only binance is supporting 'recvWindows' param when requesting balance. Other exchanges like Kraken and Bittrex (reported on telegram channel, thanks @makisasartakis) fail to get user balance. An example is Bittrex current response to get_balance :
```json
{"code":"BAD_REQUEST","detail":"Refer to the data field for specific field validation failures.","data":{"invalidRequestParameter":"recvWindow"}}
```

The fix consists to remove the unnecessary 'recvWindow' for each exchange and let the exchange tentacle adding this param if necessary (like binance tentacle is currently doing).

### Fixed
- [CCXT_Exchange] Fix unnecessary 'recvWindows' param


GuillaumeDSM (approved):
👍

PR #452, opened 2021-01-30
Handle exchange order api lag

GuillaumeDSM:


octobot_trading/personal_data/trades/trades_manager.py
```diff
@@ -53,6 +53,12 @@ def upsert_trade_instance(self, trade):
             self.trades[trade.trade_id] = trade
             self._check_trades_size()
 
+    def has_closing_trade_with_order_id(self, order_id) -> bool:
+        for trade in self.trades.values():
```
Herklos, line 57: I would have preferred that we avoided that. It can be a "big" computation if we have a lot of trades in history. I don't know how to improve that.

octobot_trading/personal_data/orders/order.py
```diff
@@ -206,13 +206,19 @@ def is_open(self):
         return self.state is None or self.state.is_open()
 
     def is_filled(self):
-        return self.state.is_filled() or (self.state.is_closed() and self.status is enums.OrderStatus.FILLED)
+        if self.state is None:
```
Herklos, line 209: How can it be possible ?

tests/personal_data/trades/test_trade_factory.py
```diff
@@ -146,6 +150,53 @@ async def test_create_trade_from_order(self):
 
         await self.stop(exchange_manager)
 
+    async def test_create_trade_from_partially_filled_order(self):
```
Herklos, line 153: :+1:

tests/personal_data/trades/test_trade_manager.py
```diff
@@ -0,0 +1,49 @@
...
+pytestmark = pytest.mark.asyncio
+
+
+@pytest.fixture
+def trade_manager_and_trader(simulated_trader):
+    _, _, trader_instance = simulated_trader
+    return personal_data.TradesManager(trader_instance), trader_instance
+
+
+def test_has_closing_trade_with_order_id(trade_manager_and_trader):
```
Herklos, line 32: :+1:

octobot_trading/personal_data/trades/trade.py
```diff
@@ -59,12 +64,14 @@ def update_from_order(self, order, creation_time=0, canceled_time=0, executed_ti
...
         self.fee = order.fee
         self.trade_id = order.order_id
+        self.origin_order_id = order.order_id
         self.simulated = order.simulated
         self.side = order.side
         self.creation_time = order.creation_time if order.creation_time > 0 else creation_time
         self.canceled_time = order.canceled_time if order.canceled_time > 0 else canceled_time
         self.executed_time = order.executed_time if order.executed_time > 0 else executed_time
         self.symbol = order.symbol
+        self.is_closing_order = order.is_closed() or order.is_filled() or order.is_cancelled()
```
Herklos, line 74: How can a trade exists if it's not linked to a closed, filled or cancelled order ?

octobot_trading/personal_data/orders/order.py
GuillaumeDSM: I has this case when creating tests, you think it can't happen in a real situation ?

octobot_trading/personal_data/trades/trade.py
GuillaumeDSM: On ccxt, a partially filled order is "open" => when we will properly handle partially filled orders, we should create trades for each actual trades of an order, even if the order is still open (in this case being partially filled)

octobot_trading/personal_data/orders/order.py
Herklos: I doesn't happened until now, and we really don't want that to happen.

octobot_trading/personal_data/trades/trade.py
Herklos: Ok I see, good idea :+1:

octobot_trading/personal_data/orders/order.py
GuillaumeDSM: ok, i'll remove this check then

Herklos (approved):

PR #461, opened 2021-02-18
WIP [Orders] add wait for order state transition

GuillaumeDSM:


octobot_trading/personal_data/orders/order_state.py
```diff
@@ -146,7 +146,8 @@ async def _refresh_order_from_exchange(self, force_synchronization=False) -> boo
                                          self.order.exchange_manager.id).get_internal_producer().
                 update_order_from_exchange(order=self.order,
                                            wait_for_refresh=True,
-                                           force_job_execution=force_synchronization))
+                                           force_job_execution=force_synchronization,
+                                           wait_for_order_state_transition=True))
```
Herklos, line 150: Why not using `wait_for_refresh` ?

GuillaumeDSM: We can but it's an AsyncJob parameter that is not passed to callbacks for now (only **kwargs are), I don't think it's a good idea to pass it to every callbacks.

Herklos:
Should we close this @GuillaumeDSM?

GuillaumeDSM:
Closed till we have a real need for this

PR #471, opened 2021-03-01
WIP [Orders] consider an order cancelled when not found on cancel request

GuillaumeDSM:
Issue is: it's currently impossible to cancel an order that is already cancelled (ccxt.OrderNotFound is raised each time preventing the cancel call to continue)
2 cases:
1. no cancel notification before cancel attempt
image
2. cancel an order about to get filled
image


GuillaumeDSM:
handled in https://github.com/Drakkar-Software/OctoBot-Trading/pull/473

PR #480, opened 2021-03-16
[TickerUpdater] Fix ignored ticker rule

Herklos:


octobot_trading/exchange_data/ticker/channel/ticker_updater.py
```diff
@@ -77,7 +77,7 @@ def _is_valid(ticker):
         try:
             # at least require close, volume and timestamp
             return ticker and \
-                all(ticker[field] is not None
+                any(ticker[field] is not None
```
GuillaumeDSM, line 80: any ? I think we want all of them, not just at least one

Herklos: Sorry I misread the comment :+1:

PR #521, opened 2021-05-21
[Portfolio] remove timestamp and datetime when present in balance result

GuillaumeDSM:


octobot_trading/constants.py
```diff
@@ -18,14 +18,18 @@
...
 CURRENT_PORTFOLIO_STRING = "Current Portfolio :"
-CONFIG_PORTFOLIO_INFO = "info"
 CONFIG_PORTFOLIO_FREE = "free"
 CONFIG_PORTFOLIO_USED = "used"
 CONFIG_PORTFOLIO_TOTAL = "total"
 CONFIG_PORTFOLIO_MARGIN = "margin"
 REAL_TRADER_STR = "[Real Trader] "
 SIMULATOR_TRADER_STR = "[Simulator] "
 
+# CCXT fields
```
Herklos, line 28: I think we should create a `ExchangeConstantsPortfolioColumns` or `ExchangeConstantsBalanceColumns` enum to store these contants.

GuillaumeDSM: You think ? To me they are not balance/portfolio specific, they are more "ccxt request" specific, maybe a `ExchangeConstantsCCXTColumns` ?

Herklos: Yes

Herklos (approved):

PR #529, opened 2021-05-25
[CryptofeedWebsocketConnector] Fix missing stop and close

Herklos:


octobot_trading/exchanges/connectors/cryptofeed_websocket_connector.py
```diff
@@ -251,3 +251,9 @@ def get_exchange_pair(self, pair) -> str:
     @classmethod
     def is_supporting_exchange(cls, exchange_candidate_name) -> bool:
         return cls.get_name() == exchange_candidate_name
+
+    async def stop(self):
+        self.client.stop()
+
+    async def close(self):
+        self.client.close()
```
GuillaumeDSM, line 259: we need both stop and close ? They should not be merged ?

Herklos: `Stop` is stops tasks but they can be recreated, `close` closes the asynchronous generators and closes the event loop. But I agree, for now we doesn't use restart feature and so we use both at the same time.

GuillaumeDSM (approved):

PR #559, opened 2021-06-27
[AbstractWebsocketConnector] Add pair and time_frame dynamically

Herklos:


octobot_trading/exchanges/connectors/cryptofeed_websocket_connector.py
```diff
@@ -64,121 +74,279 @@ def __init__(self, config: object, exchange_manager: object):
...
+        Add new time_frames to self.filtered_time_frames
+        :param time_frames: the list of time_frame to add
         """
-        cryptofeed.feedhandler.SIGNALS = ()
+        for time_frame in time_frames:
+            self._add_time_frame(time_frame)
 
-    def subscribe_candle_feed(self, exchange_symbols):
-        candle_callback = self.callback_by_feed[self.EXCHANGE_FEEDS[Feeds.CANDLE]]
+    async def reconnect(self):
```
GuillaumeDSM, line 135: I think we should name this method `reset` or something like this but not `reconnect` since it's not really reconnecting the ws

octobot_trading/exchanges/connectors/cryptofeed_websocket_connector.py
```diff
@@ -247,6 +420,9 @@ async def candle(self, feed, symbol, start, stop, interval, trades, open_price,
                                            time_frame=time_frame,
                                            symbol=symbol,
                                            candle=candle)
+            await self.push_to_channel(trading_constants.TICKER_CHANNEL,
```
GuillaumeDSM, line 423: I think it's a very good idea to create the ticker from here but we should only push it on one timeframe otherwise i'm affraid we will push it for each watched timeframe for each price change (because each candle of the same pair will receive an update)

octobot_trading/exchanges/exchange_config_data.py
```diff
@@ -13,12 +13,16 @@
 #
 #  You should have received a copy of the GNU Lesser General Public
 #  License along with this library.
+import asyncio
```
GuillaumeDSM, line 16: is this import used ?

octobot_trading/exchanges/exchange_config_data.py
```diff
@@ -53,17 +57,24 @@ def set_config_time_frame(self):  # TODO
...
+            symbol
+            for symbol in symbols
+            if self._is_valid_symbol(symbol, []) and symbol not in self.all_config_symbol_pairs
+        ]
         try:
-            return self._add_tradable_symbols(crypto_currency, symbols)
-        except Exception as e:
-            self._logger.exception(e, True, f"Fail to handle symbol update : {e}")
+            if new_valid_symbols:
+                self.all_config_symbol_pairs += new_valid_symbols
```
GuillaumeDSM, line 68: should we really update this list ? It's used to initialize evaluators etc early on, I'm not sure we should update it

octobot_trading/exchanges/exchange_config_data.py
```diff
@@ -53,17 +57,24 @@ def set_config_time_frame(self):  # TODO
     def get_shortest_time_frame(self):
         return self.traded_time_frames[-1]
 
-    async def handle_symbol_update(self, exchange: str, exchange_id: str, crypto_currency: str, symbols: list) -> tuple:
+    async def add_watched_symbols(self, symbols):
```
GuillaumeDSM, line 60: 👍

GuillaumeDSM (commented):
great !

octobot_trading/exchanges/exchange_config_data.py
Herklos: not anymore :+1:

octobot_trading/exchanges/exchange_config_data.py
Herklos: Ok, I will change it

GuillaumeDSM (approved):

PR #563, opened 2021-07-07
[AbstractWebsocketExchange] Add ignore feed when another is covering the same data

Herklos:
For example, when ticker is available in WS we don't need to subscribe to recent trade

Requires https://github.com/Drakkar-Software/OctoBot-Trading/pull/559

### DIFF
--- a/octobot_trading/exchanges/abstract_websocket_exchange.py
+++ b/octobot_trading/exchanges/abstract_websocket_exchange.py
@@ -32,6 +32,9 @@ class AbstractWebsocketExchange:
 
     INIT_REQUIRING_EXCHANGE_FEEDS = set()
 
+    # Used to ignore a feed when almost one of the corresponding feed is supported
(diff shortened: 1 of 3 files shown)


octobot_trading/exchanges/abstract_websocket_exchange.py
```diff
@@ -32,6 +32,9 @@ class AbstractWebsocketExchange:
 
     INIT_REQUIRING_EXCHANGE_FEEDS = set()
 
+    # Used to ignore a feed when almost one of the corresponding feed is supported
```
GuillaumeDSM, line 35: # Used to ignore a feed when **at least** one of the corresponding feed is supported

?

Herklos: Right :smile:

GuillaumeDSM (approved):

PR #580, opened 2021-08-05
Fix RealTime Evaluator not working on backtesting

valouvaliavlo:


GuillaumeDSM:
The issue was actually in OctoBot-Evaluators: when looking for available time frames, the wrong value was returned (see https://github.com/Drakkar-Software/OctoBot-Evaluators/pull/145) and the subscribed value in real time evaluators was also wrong (see https://github.com/Drakkar-Software/OctoBot-Tentacles/pull/520). When these will be merged, it will work.

valouvaliavlo:
👍

PR #596, opened 2021-08-25
[CCXTExchange] Add headers and options attribute

Herklos:


octobot_trading/exchanges/connectors/ccxt_exchange.py
```diff
@@ -88,6 +93,26 @@ def get_ccxt_client_login_options(self):
...
+            if self.client is not None:
+                self.client.headers[header_key] = headers_dict[header_key]
+
+    def add_options(self, options_dict):
+        """
+        Add new options to ccxt client
+        :param options_dict: the additional option keys and values as dict
+        """
+        for option_key in options_dict.keys():
+            self.options[option_key] = options_dict[option_key]
```
GuillaumeDSM, line 112:
```python
for option_key, option_value in options_dict.items():
    self.options[option_key] = option_value
```

octobot_trading/exchanges/connectors/ccxt_exchange.py
```diff
@@ -88,6 +93,26 @@ def get_ccxt_client_login_options(self):
...
+            if self.client is not None:
+                self.client.options[option_key] = options_dict[option_key]
```
GuillaumeDSM, line 114: `self.client.options[option_key] = option_value`

GuillaumeDSM: that means that `self.client` might only have some of the exchange option if it is set later ? maybe we should add every option all the time (to be sure it's always a mirror). Or upon creation all options are always passed ?

octobot_trading/exchanges/connectors/ccxt_exchange.py
```diff
@@ -88,6 +93,26 @@ def get_ccxt_client_login_options(self):
             return {'defaultType': 'margin'}
         return {'defaultType': 'spot'}
 
+    def add_headers(self, headers_dict):
+        """
+        Add new headers to ccxt client
+        :param headers_dict: the additional header keys and values as dict
+        """
+        for header_key in headers_dict.keys():
```
GuillaumeDSM, line 101: `for header_key header_value in headers_dict.idems():`

GuillaumeDSM (commented):
👍

PR #606, opened 2021-08-30
[CryptofeedWebsocketConnector] Add subscribe mark price only for watched pairs

Herklos:


octobot_trading/exchanges/abstract_websocket_exchange.py
```diff
@@ -195,10 +195,19 @@ def get_exchange_pair(self, pair):
     async def reset(self):
         raise NotImplementedError("reset is not implemented")
 
-    def add_pairs(self, pairs):
+    def add_pairs(self, pairs, watching=False):
```
GuillaumeDSM, line 198: 👍 
One quick idea: I think it would be better to name it `watching_only` or something with `only` considering that if this is true, it will not do as many things

octobot_trading/exchanges/connectors/cryptofeed_websocket_connector.py
```diff
@@ -377,37 +387,60 @@ def _subscribe_candle_feed(self):
...
                                  channels=self.channels,
                                  callbacks=self.callbacks)
             for channel in self.channels:
                 self.logger.debug(f"Subscribed to {channel}")
 
+    def _subscribe_watched_pairs_feed(self):
+        """
+        Subscribes all time frame unrelated feeds for watched pairs
+        """
+        if self.watched_pairs:
```
GuillaumeDSM, line 419: I think we should also subscribe to `self.WATCHED_PAIR_CHANNELS` for `self.filtered_pairs` otherwise regular pairs will never subscribe on `self.WATCHED_PAIR_CHANNELS`.
Or add them to `self.watched_pairs` in `_add_pair` ?

Herklos: self.WATCHED_PAIR_CHANNELS is a part of self.channels

Herklos: So filtered_pairs subscribe on self.WATCHED_PAIR_CHANNELS and many more than watched_pairs

GuillaumeDSM: Got it, great ! Should we filter `self.WATCHED_PAIR_CHANNELS` to remove unsupported ones though ?

octobot_trading/exchanges/connectors/cryptofeed_websocket_connector.py
```diff
@@ -377,37 +387,61 @@ def _subscribe_candle_feed(self):
...
                                  channels=self.channels,
                                  callbacks=self.callbacks)
             for channel in self.channels:
                 self.logger.debug(f"Subscribed to {channel}")
 
+    def _subscribe_watched_pairs_feed(self):
+        """
+        Subscribes all time frame unrelated feeds for watched pairs
+        """
+        channels = [channel for channel in self.WATCHED_PAIR_CHANNELS if self._is_supported_channel(channel)]
```
GuillaumeDSM, line 419: 👍

GuillaumeDSM (approved):
🌮

PR #627, opened 2021-09-20
[CryptofeedWebsocketConnector] Add orders callback implementation

Herklos:
Only supported (and tested) on OKEx. Don't use it for now, the cryptofeed implementation is not stable


octobot_trading/exchanges/connectors/cryptofeed_websocket_connector.py
```diff
@@ -707,7 +709,17 @@ async def order(self, order: cryptofeed_types.OrderInfo, receipt_timestamp: floa
...
-        await self.push_to_channel(trading_constants.ORDERS_CHANNEL, orders=[self.exchange.parse_order(order)])
+        await self.push_to_channel(trading_constants.ORDERS_CHANNEL, orders=[
+            personal_data.create_order_instance(trader=self.exchange_manager.trader,
+                                                order_type=self._parse_order_type(order.type),
+                                                status=self._parse_order_status(order.status),
+                                                order_id=order.id,
+                                                symbol=order.symbol,
+                                                current_price=order.price,
+                                                quantity=order.amount,
+                                                quantity_filled=order.amount - order.remaining,
```
GuillaumeDSM, line 720: is it not supposed to be a dict and it's in the ordre channel that we create the actual order ?

octobot_trading/exchanges/connectors/cryptofeed_websocket_connector.py
```diff
@@ -739,3 +786,73 @@ async def fill(self, fill: cryptofeed_types.Fill, receipt_timestamp: float):
...
+    """
+    Parsers
+    """
+
+    def _parse_order_type(self, raw_order_type):
+        """
+        :param raw_order_type: the cryptofeed unified order type
+        :return: the unified OctoBot order type
+        """
+        # TODO Migrate to switch case with python 3.10
```
GuillaumeDSM, line 799: 👍

octobot_trading/exchanges/connectors/cryptofeed_websocket_connector.py
```diff
@@ -707,7 +743,18 @@ async def order(self, order: cryptofeed_types.OrderInfo, receipt_timestamp: floa
         :param order the order object defined in cryptofeed.types.?
         :param receipt_timestamp: received timestamp
         """
-        await self.push_to_channel(trading_constants.ORDERS_CHANNEL, orders=[self.exchange.parse_order(order)])
+        await self.push_to_channel(trading_constants.ORDERS_CHANNEL, orders=[{
+            trading_enums.ExchangeConstantsOrderColumns.TYPE.value: self._parse_order_type(order.type),
```
GuillaumeDSM, line 747: should we parse these data (order type etc) ? I think what we want here is to convert it into a format similar to the order updater and it will be parsed later on in Order

octobot_trading/exchanges/connectors/cryptofeed_websocket_connector.py
```diff
@@ -429,13 +441,34 @@ def _subscribe_watched_pairs_feed(self):
...
+            feed_kwargs["candle_interval"] = candle_interval
+        self.client.add_feed(self.get_feed_name(),
+                             log_message_on_error=True,
+                             callbacks=callbacks,
+                             channels=channels,
+                             **feed_kwargs)
+        for channel in channels:
+            symbols_str = f"for {', '.join(symbols)}" if symbols else ""
+            candle_interval_str = f"on {candle_interval}" if candle_interval else ""
+            self.logger.debug(f"Subscribed to {channel} {symbols_str} {candle_interval_str}")
```
GuillaumeDSM, line 471: 👍

octobot_trading/exchanges/connectors/cryptofeed_websocket_connector.py
Herklos: I don't understand, it's exactly what we are doing here : converting cryptofeed format to a parsable format for order updater or order.

GuillaumeDSM: `_parse_order_type` returns a `trading_enums.TradeOrderType`, order updaters return a `str`, parsing of this `str` is made later in `Order`. Unless I missed something, I think we will have a different format in rest and in ws if we keep this (for the content of the order dict we provide to the order channel)

Herklos: `_parse_order_type` is returning a `str` in this merge request.

GuillaumeDSM: Oh right, my bad !

GuillaumeDSM (approved):
🎉

PR #642, opened 2021-10-20
[Future][Portfolio] Add assets

Herklos:


Herklos:
> It would be really nice to add some tests here :)

It's only additions and dict creation. The only interesting method (_ensure_update_validness) is coming from Portfolio and is already tested, I'll move its tests when migrating from portfolio dict to Assets.

PR #654, opened 2021-10-31
[TradingModes] change get_holdings_ratio to sync

GuillaumeDSM:


octobot_trading/modes/scripting_library/data/data_operators.py
```diff
@@ -0,0 +1,2 @@
+# Copyright
```
Herklos, line 1: ?

Herklos (changes requested):
:+1:

Herklos (approved):
:+1:

PR #668, opened 2021-12-28
[CryptofeedWebsocketConnector] Fix ignore ticker when KLINE not CANDLES

Herklos:


octobot_trading/exchanges/connectors/cryptofeed_websocket_connector.py
```diff
@@ -52,7 +52,7 @@ class CryptofeedWebsocketConnector(abstract_websocket.AbstractWebsocketExchange)
         # When ticker or future index is available : no need to calculate mark price from recent trades
         Feeds.TRADES: [Feeds.TICKER, Feeds.FUTURES_INDEX],
         # When candles are available : use min timeframe kline to push ticker
-        Feeds.TICKER: [Feeds.CANDLE]
+        Feeds.TICKER: [Feeds.KLINE]
```
GuillaumeDSM, line 55: we don't create ticker from candles anymore ?

Herklos: From KLINE. It's the same feed (CANDLE) for 90% of exchanges but for Kraken for example CANDLE feed doesn't provide KLINE. So we can't get mark price from kline, we have mark price only on closed candles that can be very long if using long timeframes

GuillaumeDSM: We subscribe to kline feeds in websockets ? I can't see any associated callback in `CryptofeedWebsocketConnector`

GuillaumeDSM (approved):

PR #677, opened 2022-01-20
[Asset] Add restore state when raising PortfolioNegativeError

Herklos:


octobot_trading/personal_data/portfolios/assets/future_asset.py
```diff
@@ -80,14 +95,20 @@ def update(self, total=constants.ZERO, available=constants.ZERO, position_margin
...
+        try:
+            self.initial_margin += self._ensure_update_validity(self.initial_margin, initial_margin)
+            self.position_margin = self._ensure_not_negative(self.position_margin + position_margin)
+            self.order_margin = self._ensure_not_negative(self.order_margin + available)
+            self.unrealized_pnl += unrealized_pnl
+            self.wallet_balance += self._ensure_update_validity(self.wallet_balance, total)
+            self._update_available()
+            self._update_total()
+            return True
+        except errors.PortfolioNegativeValueError as negative_value_error:
```
GuillaumeDSM, line 109: considering this try/except block is used 3 times, should we put it into a context manager ?

octobot_trading/personal_data/portfolios/assets/future_asset.py
```diff
@@ -80,14 +95,20 @@ def update(self, total=constants.ZERO, available=constants.ZERO, position_margin
...
+            self.restore(previous_asset)
```
GuillaumeDSM, line 110: Not sure I understand why restore here. It means that value that should be changed to sub 0 is just not updated and we ignore the change ?

Herklos: A raise can lead to a partial modification of Asset's attributes which is not good. What we want is that either the update is complete or that none of the attributes are modified. Otherwise we are in an unstable state where only part of the updates have been carried out. So I added a `restore` method to restore attribute values when PortfolioNegativeValueError is raised.

octobot_trading/personal_data/portfolios/asset.py
```diff
@@ -93,3 +105,15 @@ def _ensure_not_negative(self, new_value, replacement_value=constants.ZERO):
...
         return replacement_value
+
+    @contextlib.contextmanager
+    def update_or_restore(self):
+        """
+        Ensure update complete without raising PortfolioNegativeValueError else restore Asset instance's attributes
+        """
+        previous_asset = copy.copy(self)
+        try:
+            yield
```
GuillaumeDSM, line 116: 💯

GuillaumeDSM (approved):

PR #680, opened 2022-01-22
Future fixes

GuillaumeDSM:


octobot_trading/personal_data/orders/order_util.py
```diff
@@ -95,7 +96,9 @@ def check_cost(total_order_price, min_cost):
     return True
 
 
-async def get_pre_order_data(exchange_manager, symbol: str, timeout: int = None):
+async def get_pre_order_data(exchange_manager, symbol: str, timeout: int = None,
+                             portfolio_type=commons_constants.PORTFOLIO_AVAILABLE,
```
Herklos, line 100: We shouldn't use `portfolio_type` anymore. It doesn't make sense with Assets. We should migrate portfolio_type usage instead of adapting methods

octobot_trading/personal_data/portfolios/asset.py
```diff
@@ -34,6 +34,10 @@ def __str__(self):
                f"Available: {float(self.available)} | " \
                f"Total: {float(self.total)}"
 
+    def __repr__(self):
+        # __repr__ is called when a dict is turned into string (like in logs)
+        return str(self)
```
Herklos, line 39: This is not the default implementation?

octobot_trading/personal_data/positions/position.py
```diff
@@ -320,6 +320,17 @@ def _is_update_increasing_size(self, size_update):
             return size_update > constants.ZERO
         return size_update < constants.ZERO
 
+    def _is_update_decreasing_size(self, size_update):
```
Herklos, line 323: Why not using `not  _is_update_increasing_size()` ?

octobot_trading/personal_data/exchange_personal_data.py
```diff
@@ -118,8 +118,8 @@ async def handle_portfolio_update_from_funding(self, position, funding_rate,
                 changed = await self.portfolio_manager.handle_balance_update_from_funding(
                     position=position, funding_rate=funding_rate, require_exchange_update=require_exchange_update)
             except errors.PortfolioNegativeValueError:
-                self.logger.warning(True, "Not enough available balance to handle funding. Reducing position margin...")
-                await position.update(update_margin=position.value * funding_rate)
+                self.logger.warning("Not enough available balance to handle funding. Reducing position margin...")
+                await position.update(update_margin=-position.value * funding_rate)
```
Herklos, line 122: :+1: I forgot it

octobot_trading/personal_data/positions/position.py
GuillaumeDSM: `_is_update_increasing_size` is a strict comparison that returns True on a 0 size update, in this particular case, we don't want to update realized pnl and create a 0 pnl value transaction

octobot_trading/personal_data/portfolios/asset.py
GuillaumeDSM: not for dict nested elements apparently

octobot_trading/personal_data/orders/order_util.py
GuillaumeDSM: what do you see we should do in this context ?

octobot_trading/personal_data/positions/position.py
GuillaumeDSM: It's also an issue when `size_update` value is 0 as `_is_update_increasing_size` returns False while it's not actually decreasing the position

octobot_trading/personal_data/orders/order_util.py
Herklos: I think we should return the asset instance instead of asking for a `portfolio_type`.

octobot_trading/personal_data/positions/position.py
Herklos: But `size_update = 0` isn't actually increasing the position size either

GuillaumeDSM: Right, and in `_update_size` we need to know if the position is really decreasing of we will create an empty transaction, which makes no sense in y opinion

Herklos: I understand your fix but I don't understand why you didn't just replace `_is_update_increasing_size` by `_is_update_decreasing_size`?

octobot_trading/personal_data/orders/order_util.py
GuillaumeDSM: In this case, we will have to call `market_quantity = market_available / mark_price if mark_price else constants.ZERO` every time we need it after calling `get_pre_order_data` if we don't directly manipulate portfolio values but just return assets, I'm not sure it's better.

octobot_trading/personal_data/positions/position.py
GuillaumeDSM: We need both `_is_update_increasing_size` as  `_is_update_increasing_size` is also used elsewhere

Herklos: `_update_size` is the only method that use  `_is_update_increasing_size`

octobot_trading/personal_data/orders/order_util.py
Herklos: I think that we will have to think about the role of this method. It becomes a bit of a method where you put everything into it.  This problem would probably not be discussed with a method with a more precise scope in my opinion.

octobot_trading/personal_data/positions/position.py
GuillaumeDSM: Right, to me it makes sense to keep it as it's not an obvious computation and therefore having it in a dedicated method is good. Maybe you have a different idea ?

Herklos: I was just wondering if `_is_update_increasing_size` should be kept because its only use is `_update_size` and we already have the information with `not` `_is_update_decreasing_size`'s result.

GuillaumeDSM: > I was just wondering if `_is_update_increasing_size` should be kept because its only use is `_update_size` and we already have the information with `not` `_is_update_decreasing_size`'s result.

Not really because with an  `_update_size` of 0, `_is_update_increasing_size` and `_is_update_decreasing_size` should both be false, and therefore `not _is_update_decreasing_size` is not equal to `_is_update_increasing_size`

octobot_trading/personal_data/positions/types/linear_position.py
```diff
@@ -124,7 +124,8 @@ def update_average_entry_price(self, update_size, update_price):
                                           + (Update quantity * Update price)]
         """
         total_contract_value = self.size + update_size
-        self.entry_price = ((self.size * self.entry_price + update_size * update_price) /
-                            (total_contract_value if total_contract_value != constants.ZERO else constants.ONE))
+        if self._is_update_increasing_size(update_size):
```
Herklos, line 127: That's a good fix but I think it should be put in `Position.update_from_order():276` instead of here

octobot_trading/personal_data/positions/position.py
Herklos: Ok I got it :+1:

octobot_trading/personal_data/positions/types/linear_position.py
GuillaumeDSM: Right ! changing it

octobot_trading/personal_data/positions/position.py
```diff
@@ -273,7 +273,8 @@ def update_from_order(self, order):
         size_update = self._calculates_size_update_from_filled_order(order, size_to_close)
 
         # Updates position average entry price from order
-        self.update_average_entry_price(size_update, order.filled_price)
+        if self._is_update_increasing_size(size_update):
```
Herklos, line 276: :+1: Nice fix

Herklos: We can also do something like `if has_increase_position_size := self._is_update_increasing_size(size_update):` and remove `_update_size` return and `_update_size`'s `_is_update_increasing_size()` call.

Herklos: We can also do something like if has_increased_position_size := self._is_update_increasing_size(size_update): and remove _update_size return and _update_size's _is_update_increasing_size() call.

GuillaumeDSM: right

GuillaumeDSM: done in  https://github.com/Drakkar-Software/OctoBot-Trading/pull/682

PR #695, opened 2022-03-11
Scripting requirements

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Commons/pull/236
requires https://github.com/Drakkar-Software/OctoBot-Backtesting/pull/158


octobot_trading/exchanges/traders/trader.py
```diff
@@ -105,7 +105,8 @@ async def create_order(self, order, loaded: bool = False, pre_init_callback=None
         if pre_init_callback is not None:
             await pre_init_callback(new_order)
 
-        await new_order.initialize()
+        # force initialize to always create open state
+        await new_order.initialize(force=True)
```
Herklos, line 109: I don't understand this `force=True` there is no force usage in `create_order_state`

octobot_trading/personal_data/orders/order.py
```diff
@@ -314,6 +314,11 @@ def order_state_creation(self):
         except errors.InvalidOrderState as exc:
             logging.get_logger(self.get_logger_name()).exception(exc, True, f"Error when creating order state: {exc}")
 
+    async def on_pending_creation(self, is_from_exchange_data=False):
```
Herklos, line 317: 👍

octobot_trading/personal_data/orders/order.py
```diff
@@ -471,6 +477,32 @@ def update_from_raw(self, raw_order):
             reduce_only=raw_order.get(enums.ExchangeConstantsOrderColumns.REDUCE_ONLY.value, False)
         )
 
+    async def update_from_order(self, other_order):
```
Herklos, line 480: Why not using `self.update(` with `other_order` attributes ?

octobot_trading/personal_data/orders/orders_manager.py
```diff
@@ -105,26 +105,27 @@ async def upsert_order_from_raw(self, order_id, raw_order, is_from_exchange) ->
...
+        self.pending_creation_orders.append(pending_order)
 
-    def _check_pending_order(self, order):
-        for index, pending_order in enumerate(self.pending_bundled_orders):
-            if order_util.is_associated_pending_order(order, pending_order):
-                order_util.apply_pending_order(order, pending_order)
-                self.pending_bundled_orders.pop(index)
-                return True
-        return False
+    def get_and_update_pending_order(self, created_order):
```
Herklos, line 121: 👍

octobot_trading/personal_data/orders/order_state.py
```diff
@@ -77,6 +83,12 @@ async def should_be_updated(self) -> bool:
         """
         return self.order.is_self_managed()
 
+    async def change_order(self, new_order):
```
Herklos, line 86: I don't really like the word `change` here, I think `replace` would be better. What do you think?

octobot_trading/exchanges/traders/trader.py
GuillaumeDSM: right

octobot_trading/personal_data/orders/order.py
GuillaumeDSM: We could use self.update(), here I think it's better to define a real method that will update all the necessary attributes in one place when we have an existing order instead of making a huge update() for which we will need to specify each attribute every other time we will need to udpate an order from another one.
This way we know that it's only in this method that we can update an order from another order thazt is already in octobot

octobot_trading/personal_data/orders/order_state.py
GuillaumeDSM: right

octobot_trading/exchanges/traders/trader.py
GuillaumeDSM: actually force was to force initialization even if already initialized. However as it's unclear, I changed directly the `is_initialized attribute`

Herklos:
This is only an aggregation of merged PR right?

GuillaumeDSM:
> This is only an aggregation of merged PR right?

all commits that are merge or [TM] are yep, others are real

octobot_trading/personal_data/orders/order.py
```diff
@@ -418,8 +430,12 @@ async def default_exchange_update_order_status(self):
     def generate_executed_time(self):
         return self.exchange_manager.exchange.get_exchange_current_time()
 
+    def is_counted_in_available_funds(self):
```
Herklos, line 433: 👍

octobot_trading/personal_data/orders/types/limit/stop_loss_order.py
```diff
@@ -22,9 +22,16 @@ def __init__(self, trader, side=enums.TradeOrderSide.SELL):
         super().__init__(trader, side)
         self.trigger_above = self.side is enums.TradeOrderSide.BUY
 
+    def is_counted_in_available_funds(self):
+        return False
+
+    def _filled_maker_or_taker(self):
+        # Creates a market order when filled, which is taker
```
Herklos, line 29: 👍

octobot_trading/personal_data/orders/types/market/market_order.py
```diff
@@ -13,20 +13,19 @@
...
 class MarketOrder(order_class.Order):
     async def update_order_status(self, force_refresh=False):
         if self.trader.simulate:
-            asyncio.create_task(self.on_fill(force_fill=True))
-            # In trading simulation wait for the next asyncio loop iteration to ensure this order status
-            # is updated before leaving this method
-            await asyncio_tools.wait_asyncio_next_cycle()
+            # TODO: ensure no issue un not running it in task anymore
+            await self.on_fill(force_fill=True)
+            # asyncio.create_task(self.on_fill(force_fill=True))
```
Herklos, line 25: Should we merge it? Are we sure?

octobot_trading/personal_data/portfolios/portfolio.py
```diff
@@ -257,14 +275,23 @@ def log_portfolio_update_from_order(self, order):
...
+
 
 def _should_update_available(order):
     """
     Check if the order has impact on availability
     :param order: The order to check
     :return: True if the order should update available portfolio
     """
-    return not order.is_self_managed()
+    return order.is_counted_in_available_funds()
```
Herklos, line 294: 👍

octobot_trading/personal_data/portfolios/portfolio_manager.py
```diff
@@ -89,13 +89,29 @@ async def handle_balance_update_from_funding(self, position, funding_rate, requi
             return await self._refresh_real_trader_portfolio()
         return False
 
+    async def handle_balance_update_from_withdrawal(self, amount, currency) -> bool:
```
Herklos, line 92: 👍

octobot_trading/personal_data/portfolios/portfolio_util.py
```diff
@@ -45,3 +45,26 @@ def portfolio_to_float(portfolio):
                 "total": float(symbol_balance.total)
             }
     return float_portfolio
+
+
+def get_draw_down(exchange_manager):
```
Herklos, line 50: nice feature!

octobot_trading/personal_data/portfolios/portfolio_util.py
```diff
@@ -45,3 +45,26 @@ def portfolio_to_float(portfolio):
...
+    :param exchange_manager:
+    :return:
+    """
+    draw_down = 0
+    if exchange_manager.is_future:
+        origin_portfolio = portfolio_to_float(
+            exchange_manager.exchange_personal_data.portfolio_manager.portfolio_value_holder.origin_portfolio.portfolio
+        )
+        portfolio_history = [
+            origin_portfolio[exchange_manager.exchange_personal_data.portfolio_manager.reference_market]["total"]]
```
Herklos, line 62: Shouldn't we use contants for these portfolio history keys?

octobot_trading/personal_data/trades/trades_util.py
```diff
@@ -0,0 +1,51 @@
...
+import octobot_trading.personal_data as personal_data
+
+
+_LOSING_ORDER_TYPES = [
+    trading_enums.TradeOrderType.STOP_LOSS,
+    trading_enums.TradeOrderType.STOP_LOSS_LIMIT,
+]
+
+
+def compute_win_rate(exchange_manager):
```
Herklos, line 27: Cool 👌

tests/modes/script_keywords/__init__.py
```diff
@@ -0,0 +1 @@
+# Copyright
```
Herklos, line 1: Only the first line?

tests/signals/test_signal_builder.py
```diff
@@ -0,0 +1,326 @@
...
+    assert trading_signal.description == "signal description"
+    assert trading_signal.state == enums.EvaluatorStates.SHORT.value
+    assert trading_signal.orders == []
+
+    signal_builder.orders = [None]
+    assert signal_builder.build().orders == [None]
+
+
+def test_reset(signal_builder):
+    signal_builder.orders = ["hi"]
```
Herklos, line 58: 😄

octobot_trading/api/portfolio.py
```diff
@@ -57,3 +58,15 @@ async def refresh_real_trader_portfolio(exchange_manager) -> bool:
     return await exchange_channel.get_chan(octobot_trading.constants.BALANCE_CHANNEL, exchange_manager.id). \
         get_internal_producer(). \
         refresh_real_trader_portfolio(True)
+
+
+def get_portfolio_amounts(portfolio,
+                          portfolio_type=octobot_commons.constants.PORTFOLIO_TOTAL):
```
Herklos, line 64: We are not using PORTFOLIO_TOTAL anymore, I don't understand this method

octobot_trading/exchange_data/ohlcv/candles_manager.py
```diff
@@ -98,6 +104,24 @@ def get_symbol_prices(self, limit=-1):
             enums.PriceIndexes.IND_PRICE_TIME.value: self.get_symbol_time_candles(limit)
         }
 
+    def get_candles(self, limit=-1):
```
Herklos, line 107: Where is it used?

Herklos (commented):
Awesome work ! 💯

octobot_trading/enums.py
```diff
@@ -466,3 +468,159 @@ class SubAccountColumns(enum.Enum):
 class ExchangeWrapperLibs(enum.Enum):
     ASYNC_CCXT = "async_ccxt"
     CCXT = "ccxt"
+
+
+class PlotAttributes(enum.Enum):
```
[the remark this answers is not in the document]
GuillaumeDSM: Right, its more like a historical thing. Those are used to save data into run databases (orders and other things). They have to be accessible from trading and tentacles. We could but this in a dedicated file in `commons/databases`, what do you think ?

octobot_trading/exchanges/exchange_manager.py
```diff
@@ -197,6 +202,13 @@ def get_exchange_name(self):
     def get_currently_handled_pair_with_time_frame(self):
         return len(self.exchange_config.traded_symbol_pairs) * len(self.exchange_config.traded_time_frames)
 
+    def ensure_reachability(self):
+        if self.exchange.is_unreachable:
+            current_time = self.exchange.get_exchange_current_time()
```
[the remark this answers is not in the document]
GuillaumeDSM: The purpose of this method is not to check if the exchange is reachable by making a request to it but just to check the  `self.exchange.is_unreachable` attribute. This line is just used to build the error message, it's not checking anything.
Do you think we should add docs to explain that this is only relevant in backtesting for now ? (the live trading part is not coded, it's always considered reachable)

octobot_trading/exchange_data/ticker/channel/ticker_updater_simulator.py
```diff
@@ -67,9 +71,10 @@ async def _ticker_from_ohlcv_callback(self, exchange: str, exchange_id: str,
                                           cryptocurrency: str, symbol: str, time_frame, candle):
         if self.ticker_time_frame == time_frame and candle:
             last_candle_timestamp = candle[common_enums.PriceIndexes.IND_PRICE_TIME.value]
-            if last_candle_timestamp > self.last_timestamp_pushed_by_symbol[symbol]:
-                self.last_timestamp_pushed_by_symbol[symbol] = last_candle_timestamp
-                ticker = self._generate_ticker_from_candle(candle, symbol, last_candle_timestamp)
+            ticker_timestamp = last_candle_timestamp + self.ticker_time_frame_seconds
+            if ticker_timestamp > self.last_timestamp_pushed_by_symbol[symbol]:
+                self.last_timestamp_pushed_by_symbol[symbol] = ticker_timestamp
+                ticker = self._generate_ticker_from_candle(candle, symbol, ticker_timestamp)
```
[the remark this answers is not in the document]
GuillaumeDSM: on exchange data, the timestamp value of a ticker is the current time, not the candle open time, here we do the same (previously we were using `last_candle_timestamp` which is the open time of the last candle

octobot_trading/enums.py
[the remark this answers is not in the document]
Herklos: Yes 👍

octobot_trading/exchanges/exchange_manager.py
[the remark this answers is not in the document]
Herklos: Oh ok, I think we should add a doc here because I thought it was only used in live trading

octobot_trading/exchanges/connectors/cryptofeed_websocket_connector.py
```diff
@@ -684,7 +684,7 @@ async def candle(self, candle_data: cryptofeed_types.Candle, receipt_timestamp:
             Ectc.LAST.value: float(candle_data.close),
             Ectc.PREVIOUS_CLOSE.value: None,
             Ectc.BASE_VOLUME.value: float(candle_data.volume),
-            Ectc.TIMESTAMP.value: candle_data.timestamp,
+            Ectc.TIMESTAMP.value: self.exchange.get_exchange_current_time(),
```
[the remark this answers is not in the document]
GuillaumeDSM: in ccxt ticker, the timestamp value is the current time, not the open time of the candle (same as above)

octobot_trading/exchanges/connectors/exchange_simulator.py
```diff
@@ -110,6 +111,16 @@ def get_available_time_frames(self):
...
 
+    def get_backtesting_data_files(self):
+        return [backtesting_api.get_data_file_path(importer) for importer in self.exchange_importers]
+
+    def get_backtesting_data_file(self, symbol, time_frame):
+        for importer in self.exchange_importers:
+            if symbol in backtesting_api.get_available_symbols(importer) and \
+                    time_frame in backtesting_api.get_available_time_frames(importer):
+                return backtesting_api.get_data_file_path(importer)
+        return None
```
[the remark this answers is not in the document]
GuillaumeDSM: Right, it's not really the exchange's job to know this. (I guess) I put it there because it's the exchange that contains its `exchange_importers` and it was easier this way. We could add an octobot-backtesting api to do the same thing with a list of importers and call it directly from the trading api (from which `get_backtesting_data_file` is called)

octobot_trading/exchanges/exchange_manager.py
```diff
@@ -117,10 +119,13 @@ async def register_trader(self, trader):
         await self.exchange_personal_data.initialize()
         await self.exchange_config.initialize()
 
-    def load_constants(self):
-        self._load_config_symbols_and_time_frames()
-        self.exchange_config.set_config_time_frame()
-        self.exchange_config.set_config_traded_pairs()
+    def load_constants(self, is_simulated_exchange):
+        if not is_simulated_exchange:
```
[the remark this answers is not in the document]
GuillaumeDSM: I guess you mean `exchange_manager.is_backtesting`. Right, changing it

octobot_trading/exchanges/traders/trader.py
```diff
@@ -437,6 +454,17 @@ async def close_position(self, position, limit_price=None, timeout=1):
                     await self.create_order(current_order))
         return created_orders
 
+    async def withdraw(self, amount, currency):
```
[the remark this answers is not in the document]
GuillaumeDSM: Where do you think it will fit more ?

octobot_trading/exchanges/util/exchange_util.py
```diff
@@ -152,3 +154,28 @@ async def is_compatible_account(exchange_name: str, exchange_config: dict, tenta
...
+        if candles:
+            start_time = candles[-1][common_enums.PriceIndexes.IND_PRICE_TIME.value]
+            while start_time > end_time and candles:
+                start_time = candles.pop(-1)[common_enums.PriceIndexes.IND_PRICE_TIME.value]
+                reached_max = True
+            if candles:
+                local_exchange_manager.exchange.uniformize_candles_if_necessary(candles)
+                yield candles
+                # avoid fetching the last element twice
+                start_time += 1
```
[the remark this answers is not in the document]
GuillaumeDSM: start_time is updated each time by `start_time = candles.pop(-1)[common_enums.PriceIndexes.IND_PRICE_TIME.value]`. We just add a second to it to make sure we fetch the next candle batch starting from the next candle and dont fetch the last candle of the previous batch twice (select on < vs <= situation)

octobot_trading/exchanges/traders/trader.py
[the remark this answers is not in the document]
Herklos: Hum... that's a good question. When I read this I thought about the fact that we only had order related methods in trader. But orders related operations are the only ones we carry out, so it makes sense that we don't have any yet that concern the pf. I don't know if we need another component to perform pf operations or if we should keep using trader for it

octobot_trading/exchanges/exchange_manager.py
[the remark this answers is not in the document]
Herklos: Yes

octobot_trading/exchanges/connectors/exchange_simulator.py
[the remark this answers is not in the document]
Herklos: Yes it would be perfect like that

octobot_trading/exchanges/connectors/cryptofeed_websocket_connector.py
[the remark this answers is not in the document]
Herklos: Ok

octobot_trading/exchange_data/ticker/channel/ticker_updater_simulator.py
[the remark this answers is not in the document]
Herklos: Ok 👍

octobot_trading/modes/abstract_trading_mode.py
```diff
@@ -137,8 +154,19 @@ async def stop(self) -> None:
...
 
+    async def close_caches(self, reset_cache_db_ids=False):
+        for tentacle_name in [self.get_name()] + [evaluator.get_name() for evaluator in self.called_nested_evaluators]:
+            await databases.CacheManager().close_cache(
+                tentacle_name,
+                self.exchange_manager.exchange_name,
+                common_constants.UNPROVIDED_CACHE_IDENTIFIER if self.get_is_symbol_wildcard() else self.symbol,
+                common_constants.UNPROVIDED_CACHE_IDENTIFIER if self.get_is_time_frame_wildcard() else self.time_frame,
+                reset_cache_db_ids=reset_cache_db_ids
+            )
```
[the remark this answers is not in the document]
GuillaumeDSM: It seems that this method is also reconstructing cache names, it don't think we should duplicate this code that is probably also used somewhere else
> right, I'm updating this part to use the same code as in evaluators

I don't think this is a good idea to make a dependance between the trading mode close and evaluators (they should be completely independent)
> here the original need is to close evaluator caches that might have been opened in nested evaluator calls while the evaluators themselves are not activated as "standalone evaluators" they will never be stopped (because never started) and therefore their cache won't be closed. Maybe we should but this into an octobot-evaluators util function ?

octobot_trading/modes/abstract_trading_mode.py
```diff
@@ -233,3 +261,10 @@ def get_required_strategies_count(cls, config):
         if constants.TRADING_MODE_REQUIRED_STRATEGIES_MIN_COUNT in config:
             min_strategies_count = config[constants.TRADING_MODE_REQUIRED_STRATEGIES_MIN_COUNT]
         return min_strategies_count
+
+    @classmethod
+    def get_required_candles_count(cls, tentacles_setup_config: tm_configuration.TentaclesSetupConfiguration):
+        return tentacles_manager_api.get_tentacle_config(tentacles_setup_config, cls).get(
+            constants.CONFIG_CANDLES_HISTORY_SIZE_KEY,
+            common_constants.DEFAULT_IGNORED_VALUE
+        )
```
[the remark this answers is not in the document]
GuillaumeDSM: you mean it should be in a `util` file ?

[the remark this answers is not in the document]
Herklos: Yes

octobot_trading/modes/channel/abstract_mode_producer.py
```diff
@@ -13,18 +13,33 @@
...
+import octobot_trading.errors as errors
 
 
 class AbstractTradingModeProducer(modes_channel.ModeChannelProducer):
+    TOPIC_TO_CHANNEL_NAME = {
+        common_enums.ActivationTopics.FULL_CANDLES.value:
+            channels_name.OctoBotTradingChannelsName.OHLCV_CHANNEL.value,
+        common_enums.ActivationTopics.IN_CONSTRUCTION_CANDLES.value:
+            channels_name.OctoBotTradingChannelsName.KLINE_CHANNEL.value,
+        common_enums.ActivationTopics.EVALUATORS.value:
```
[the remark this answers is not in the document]
GuillaumeDSM: It both yep, here the idea is to have something that in meaningful for the user, hense these translations: 
image

However we could name the enumerate itself `EVALUATOR_CYCLE` or something like this ?

octobot_trading/modes/channel/abstract_mode_producer.py
```diff
@@ -131,7 +218,15 @@ def flush(self) -> None:
         """
         self.trading_mode = None
         self.exchange_manager = None
-        self.matrix_consumer = None
+        self.consumers = []
```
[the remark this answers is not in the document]
GuillaumeDSM: 😮

octobot_trading/modes/channel/abstract_mode_producer.py
```diff
@@ -13,18 +13,33 @@
...
+import octobot_trading.storage as storage
+import octobot_trading.errors as errors
 
 
 class AbstractTradingModeProducer(modes_channel.ModeChannelProducer):
+    TOPIC_TO_CHANNEL_NAME = {
+        common_enums.ActivationTopics.FULL_CANDLES.value:
+            channels_name.OctoBotTradingChannelsName.OHLCV_CHANNEL.value,
+        common_enums.ActivationTopics.IN_CONSTRUCTION_CANDLES.value:
+            channels_name.OctoBotTradingChannelsName.KLINE_CHANNEL.value,
```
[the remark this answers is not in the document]
GuillaumeDSM: By default the "after strategies" setting is enabled therefore they will not receive candle updated unless they ask for it in their config (and they would need to implement the callback

octobot_trading/modes/channel/abstract_mode_producer.py
```diff
@@ -162,13 +258,20 @@ async def finalize(self, exchange_name: str,
...
             await self.set_final_eval(matrix_id=matrix_id,
                                       cryptocurrency=cryptocurrency,
                                       symbol=symbol,
                                       time_frame=time_frame)
+
+    @contextlib.contextmanager
+    def trading_mode_trigger(self):
+        try:
+            yield
+        except errors.UnreachableExchange as e:
```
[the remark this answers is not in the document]
GuillaumeDSM: For now it's only a backtesting thing this `errors.UnreachableExchange` therefore I would say not for now

octobot_trading/modes/channel/abstract_mode_producer.py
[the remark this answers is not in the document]
Herklos: Ok I understand. I would prefer something like `EVALUATOR_CYCLE`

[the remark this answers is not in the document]
Herklos: or evaluation cycle

octobot_trading/modes/channel/abstract_mode_producer.py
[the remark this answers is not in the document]
Herklos: It seemed to me to have read that the subscribe to OHLCV was made by default

octobot_trading/modes/channel/abstract_mode_producer.py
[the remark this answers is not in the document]
Herklos: Right

octobot_trading/personal_data/orders/types/market/market_order.py
GuillaumeDSM: I didn't spot any issue regarding this, I think we can

octobot_trading/api/portfolio.py
GuillaumeDSM: It's not used anymore, I'm removing it

octobot_trading/exchange_data/ohlcv/candles_manager.py
GuillaumeDSM: in `AbstractScriptedTradingMode#_wait_for_symbol_init` to call candles callback for the init call when enough data have been fetched

octobot_trading/enums.py
```diff
@@ -123,6 +123,11 @@ class TransactionType(enum.Enum):
     TRANSFER = "transfer"
 
 
+class BlockchainTypes(enum.Enum):
```
Herklos, line 126: 👍

octobot_trading/exchanges/exchange_config_data.py
```diff
@@ -56,7 +56,7 @@ def __init__(self, exchange_manager):
         self.real_time_time_frames = []
 
         # number of required historical candles
-        self.required_historical_candles_count = -1
+        self.required_historical_candles_count = constants.DEFAULT_IGNORED_VALUE
```
Herklos, line 59: 👍

octobot_trading/modes/channel/abstract_mode_producer.py
```diff
@@ -309,11 +310,17 @@ async def cancel_symbol_open_orders(self, symbol) -> bool:
...
-    def writers(self):
+    def all_databases(self):
         provider = storage.RunDatabasesProvider.instance()
-        return provider.get_run_db(self.trading_mode.bot_id), \
-            provider.get_orders_db(self.trading_mode.bot_id, self.exchange_name), \
-            provider.get_trades_db(self.trading_mode.bot_id, self.exchange_name), \
-            provider.get_transactions_db(self.trading_mode.bot_id, self.exchange_name), \
-            provider.get_symbol_db(self.trading_mode.bot_id, self.exchange_name, self.trading_mode.symbol) \
-                if self.trading_mode.symbol else None
+        return {
```
Herklos, line 315: 👍

octobot_trading/personal_data/portfolios/portfolio_util.py
```diff
@@ -41,25 +42,26 @@ def portfolio_to_float(portfolio):
...
 
 def get_draw_down(exchange_manager):
     """
     Draw down is the lowest portfolio value in % ever reached during a run
-    :param exchange_manager:
-    :return:
+    :return: the Draw down value
     """
-    draw_down = 0
+    draw_down = 0.0
```
Herklos, line 57: Not decimal ?

octobot_trading/personal_data/portfolios/portfolio_util.py
```diff
@@ -69,4 +70,103 @@ def get_draw_down(exchange_manager):
...
+                end_value = end_balance
+
+            def get_ideal_value(linear_growth, adj1, adj2):
+                return ((linear_growth + adj1) ** pw) * adj2
+
+            x = [start_time, end_time]
+            y = [start_balance, end_value]
+
+            pw = 15
+            A = numpy.exp(numpy.log(y[0] / y[1]) / pw)
```
Herklos, line 130: 😕

octobot_trading/exchange_data/ohlcv/candles_manager.py
Herklos: ok

octobot_trading/modes/channel/abstract_mode_producer.py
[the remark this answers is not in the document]
GuillaumeDSM: I updated this comment to be clearer 
image

octobot_trading/personal_data/portfolios/portfolio_util.py
GuillaumeDSM: updated

octobot_trading/modes/abstract_trading_mode.py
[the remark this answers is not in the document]
Herklos: Maybe. The problem here could also be the way we use standalone Evaluator. In my opinion there is no reason why something else than thé Evaluator itself should care about it's own cache

Herklos (changes requested):

octobot_trading/exchange_data/prices/price_events_manager.py
```diff
@@ -34,6 +34,7 @@ class PriceEventsManager:
     def __init__(self):
         self.logger = logging.get_logger(self.__class__.__name__)
         self.events = []
+        self._last_recent_trades = []
```
Herklos, line 37: I don't think it's a good idea to store recent trades because the price is calculated from many sources not only recent trades. Price Manager triés to abstract the way we manage price, I think we shouldn't break this abstraction. To obtain the same goal, we should store the last X prices within a predefined time (it will contain recent trades an other sources prices) What do you think?

octobot_trading/exchange_data/prices/price_events_manager.py
```diff
@@ -75,6 +77,23 @@ def add_event(self, price, timestamp, trigger_above):
         self.events.append(price_event_tuple)
         return price_event_tuple[PriceEventsManager.PRICE_EVENT_INDEX]
 
+    def is_triggered_by_last_recent_trades(self, price, timestamp, trigger_above):
```
Herklos, line 80: I think we shouldn't trigger this manually, we should check each time we try to create a new price event. It should be abstracted for the price event client.

GuillaumeDSM:
Maybe. The problem here could also be the way we use standalone Evaluator. In my opinion there is no reason why something else than thé Evaluator itself should care about it's own cache

> I agree on the principle that it's weird that a trading mode is closing an evaluator cache. The main issue for now is that we don't keep track of evaluator caches open during nested calls (from trading modes or evaluators). Maybe we could add something to the cache manager to know that it should close X Y and Z caches and add a close() method that does that when the bot is stopping ?

octobot_trading/exchange_data/prices/price_events_manager.py
GuillaumeDSM: Agree, I'll draft something

octobot_trading/exchange_data/prices/price_events_manager.py
GuillaumeDSM: Currently when adding an event, we create this event and return it. We could check if this event would be filled instantly when creating it and if so, set it instantly. Then the client would have to check if the event is set or not before creating a waiting task or proceeding as now when instantly filled.
What do you think about this approach ?

octobot_trading/exchange_data/prices/price_events_manager.py
GuillaumeDSM: idea pushed :)

octobot_trading/exchange_data/prices/price_events_manager.py
```diff
@@ -77,17 +89,17 @@ def add_event(self, price, timestamp, trigger_above):
         self.events.append(price_event_tuple)
         return price_event_tuple[PriceEventsManager.PRICE_EVENT_INDEX]
 
-    def is_triggered_by_last_recent_trades(self, price, timestamp, trigger_above):
+    def is_triggered_by_last_recent_prices(self, price, timestamp, trigger_above):
```
Herklos, line 92: 👍

octobot_trading/exchange_data/prices/price_events_manager.py
```diff
@@ -77,17 +89,17 @@ def add_event(self, price, timestamp, trigger_above):
...
         :param timestamp: the timestamp to wait for
         :param trigger_above: True if waiting for an upper price
         :return: True if it would be triggered
         """
-        for recent_trade in self._last_recent_trades:
-            trade_price = decimal.Decimal(str(recent_trade[ECOC.PRICE.value]))
-            if timestamp <= recent_trade[ECOC.TIMESTAMP.value] and (
+        for recent_price in self._last_recent_prices:
+            trade_price = recent_price[self.PRICE_KEY]
+            if timestamp <= recent_price[self.TIME_KEY] and (
```
Herklos, line 102: 👍

octobot_trading/exchange_data/prices/price_events_manager.py
Herklos: it's a good question. I think we need to abstract this internal mechanism because, in my opinion, a client shouldn't have to check before asking. A client just wants to "wait for price hit". If we set the event instantly the client has to check before waiting that the event is not set right? I think it's better, it seems logical to me

Herklos:
> Maybe. The problem here could also be the way we use standalone Evaluator. In my opinion there is no reason why something else than the Evaluator itself should care about its own cache
> 
> > I agree on the principle that it's weird that a trading mode is closing an evaluator cache. The main issue for now is that we don't keep track of evaluator caches open during nested calls (from trading modes or evaluators). Maybe we could add something to the cache manager to know that it should close X Y and Z caches and add a close() method that does that when the bot is stopping ?

With this mechanism, the only way to close nested calls evaluator cache is to stop the bot?

octobot_trading/personal_data/portfolios/portfolio_util.py
techfreaque: im not fully sure what those lines do. I got it from here: https://stackoverflow.com/a/33190889

GuillaumeDSM:
> > Maybe. The problem here could also be the way we use standalone Evaluator. In my opinion there is no reason why something else than the Evaluator itself should care about its own cache
> > > I agree on the principle that it's weird that a trading mode is closing an evaluator cache. The main issue for now is that we don't keep track of evaluator caches open during nested calls (from trading modes or evaluators). Maybe we could add something to the cache manager to know that it should close X Y and Z caches and add a close() method that does that when the bot is stopping ?
> 
> With this mechanism, the only way to close nested calls evaluator cache is to stop the bot?

selected solution: close all caches at the end of backtestings

Herklos (approved):

octobot_trading/modes/channel/abstract_mode_producer.py
```diff
@@ -83,46 +107,109 @@ async def start(self) -> None:
...
+                    priority_level=self.priority_level,
+                    cryptocurrency=currency_filter,
+                    symbol=symbol_filter,
+                    time_frame=self.time_frame_filter
+                )
+                self.trading_consumers.append(
+                    (consumer, registration_topic)
+                )
+                if registration_topic == channels_name.OctoBotTradingChannelsName.OHLCV_CHANNEL.value:
+                    self.trading_mode.is_triggered_after_candle_close = True
```
Herklos, line 168: Could we create a method to separate this code from filter definition (to simplify this method), like `_subscribe_to_registration_topic`?

octobot_trading/signals/channel/signal_producer.py
```diff
@@ -0,0 +1,50 @@
...
+class RemoteTradingSignalProducer(signals_channel.RemoteTradingSignalChannelProducer):
+    def __init__(self, channel, authenticator, bot_id):
+        super().__init__(channel)
+        # the trading mode instance logger
+        self.logger = logging.get_logger(self.__class__.__name__)
+
+        # Define trading modes default consumer priority level
+        self.priority_level: int = channel_enums.ChannelConsumerPriorityLevels.MEDIUM.value
+
+        self.authenticator = authenticator
```
Herklos, line 32: Should we also flush `authenticator`?

octobot_trading/signals/trading_signals_emitter.py
```diff
@@ -0,0 +1,25 @@
...
+#  License along with this library.
+import octobot_commons.enums as commons_enums
+import octobot_commons.authentication as authentication
+
+
+async def emit_remote_trading_signal(signal, product_slug):
+    await authentication.Authenticator.instance().send(
+        signal.to_dict(),
+        commons_enums.CommunityChannelTypes.SIGNAL.value,
+        product_slug
```
Herklos, line 24:
```suggestion
        identifier
```

octobot_trading/personal_data/portfolios/portfolio_util.py
Herklos: @techfreaque did you test lines? How can we ensure it is working?

octobot_trading/signals/channel/signal_producer.py
```diff
@@ -16,25 +16,25 @@
...
         # Define trading modes default consumer priority level
         self.priority_level: int = channel_enums.ChannelConsumerPriorityLevels.MEDIUM.value
 
-        self.authenticator = authenticator
         self.bot_id = bot_id
 
     async def subscribe_to_product_feed(self, feed_id):
-        await self.authenticator.register_feed_callback(commons_enums.CommunityChannelTypes.SIGNAL, self.on_new_signal,
-                                                        identifier=feed_id)
+        await authentication.Authenticator.instance().register_feed_callback(commons_enums.CommunityChannelTypes.SIGNAL, self.on_new_signal,
```
Herklos, line 36: 👍

Herklos (approved):

GuillaumeDSM:
Merge ? 🎉

Herklos (approved):
Yes ;) 🎉

PR #705, opened 2022-06-04
Fix typing

GuillaumeDSM:


octobot_trading/personal_data/portfolios/history/historical_asset_value_factory.pxd
```diff
@@ -16,4 +16,4 @@
 #  License along with this library.
 
 
-cpdef object create_historical_asset_value_from_dict(object historical_asset_class, dict asset_dict)
+cpdef object create_historical_asset_value_from_dict(object historical_asset_class, object asset_dict)
```
Herklos, line 19: Should we keep this name while it's no longer a dict?

Herklos (approved):

PR #714, opened 2022-07-15
[Exchanges] fix exchanges cdef methods visibility

GuillaumeDSM:


octobot_trading/exchanges/implementations/future_ccxt_exchange.pxd
```diff
@@ -18,4 +18,4 @@ cimport octobot_trading.exchanges.types as exchanges_types
 
 cdef class FutureCCXTExchange(exchanges_types.FutureExchange):
     # cpdef str get_default_type(self)   generating a segfault on exchange that both extend spot and future exchanges
-    cdef str _get_pair_market_type(self, str pair)
+    cpdef str _get_pair_market_type(self, str pair)
```
Herklos, line 21: Should we `cpdef`a `_`method?

octobot_trading/exchanges/abstract_exchange.pxd
```diff
@@ -78,8 +78,8 @@ cdef class AbstractExchange(util.Initializable):
     cpdef object get_uniformized_timestamp(self, object candle_or_candles)
     cpdef long long get_candle_since_timestamp(self, object time_frame, int count)
 
-    cdef object _uniformize_candles_timestamps(self, list candles)
-    cdef void _uniformize_candle_timestamps(self, list candle)
+    cpdef object _uniformize_candles_timestamps(self, list candles)
+    cpdef void _uniformize_candle_timestamps(self, list candle)
```
Herklos, line 82: Should we `cpdef`a `_`method?

GuillaumeDSM: This one we don't use yet, we can reverse it if you think it's better. To me this is the kind of method that has pretty high chances to be called from an exchange tentacle as well (I checked all the exchanges package cdef methods and those ones are the only ones that I think make sense to keep protected and not private)

octobot_trading/exchanges/implementations/future_ccxt_exchange.pxd
GuillaumeDSM: right !

Herklos (approved):

PR #726, opened 2022-08-31
Fix args

GuillaumeDSM:
fixes 
image

=> I don't really understand why sometimes it works and sometimes it doesnt, it probably has to do with something about cython. Anyway I think it's better not to use kw args when function args are positional and not keyword ones.


Herklos (approved):
👍

PR #769, opened 2022-11-14
Fix trailing stop

Phodia:


valouvaliavlo (approved):

GuillaumeDSM:
Thanks for this PR ! 
Did you have a chance to test it both on simulated and real trading ?

valouvaliavlo:
@GuillaumeDSM 
Only on simulation and backtesting.
For real trading, it's same as stop_loss

Phodia:
Can you approve to launch workflow ?

Herklos:
Can you add a test to ensure it's fixed please? Because it seems that we doesn't have a test for that yet.

PR #773, opened 2022-11-28
[API] add get_exchange_type

GuillaumeDSM:


octobot_trading/personal_data/orders/order_state.py
```diff
@@ -96,11 +96,15 @@ async def _synchronize_with_exchange(self, force_synchronization: bool = False)
...
-                                           force_job_execution=force_synchronization))
+        try:
+            self.ensure_not_cleared(self.order)
+            return (await exchange_channel.get_chan(octobot_trading.constants.ORDERS_CHANNEL,
+                                                    self.order.exchange_manager.id).get_internal_producer().
+                    update_order_from_exchange(order=self.order,
+                                               wait_for_refresh=True,
+                                               force_job_execution=force_synchronization))
+        except octobot_trading.errors.InvalidOrderState:
+            self.get_logger().debug(f"Skipping exchange synchronisation as order has already been closed.")
```
Herklos, line 107: Why didn't it happen before?

GuillaumeDSM: I think it did and we didn't realize it. it was just producing an error, it had no impact on the bot as it was happening when the order updater tries to refresh an order that is (simultaneously) being refreshed from another source. In this case we enter synchronize when the order is alredy closed (because syncrhonize is called from an async task that is not started rightaway and in the time being the order might get closed)

GuillaumeDSM: (and refreshing a closed order has no impact, it just creates an error as there is no associated exchange manager)

Herklos (approved):

PR #782, opened 2022-12-10
[Exchange] add parsers base

techfreaque:
I'll close this PR and start a new one once I rebased my changes. As this PR history will not be relevant anymore.

as discussed the base for the parsers . If you want to try them use the branch  make_exchanges_more_generic


octobot_trading/enums.py
```diff
@@ -217,6 +223,7 @@ class ExchangeOrderCCXTColumns(enum.Enum):
     REMAINING = "remaining"
     STATUS = "status"
     FEE = "fee"
+    FEES = "fees"
```
Herklos, line 226: As my first comment, I don't think we should have both fee and fees

octobot_trading/enums.py
```diff
@@ -364,12 +373,14 @@ class ExchangeConstantsOrderColumns(enum.Enum):
     SYMBOL = "symbol"
     QUANTITY_CURRENCY = "quantity_currency"
     TYPE = "type"
+    OCTOBOT_ORDER_TYPE = "octobot_order_type"
```
Herklos, line 376: I don't understand why we should store this value here

octobot_trading/enums.py
```diff
@@ -381,9 +392,22 @@ class ExchangeConstantsOrderColumns(enum.Enum):
     REDUCE_ONLY = "reduceOnly"
     STOP_PRICE = "stopPrice"
     TRIGGER_ABOVE = "triggerAbove"
+    POST_ONLY = "post_only"
```
Herklos, line 395: Are we supporting it now?

octobot_trading/enums.py
```diff
@@ -392,12 +416,12 @@ class ExchangeConstantsPositionColumns(enum.Enum):
...
-    REALISED_PNL = "realised_pnl"
+    UNREALIZED_PNL = "unrealized_pnl"  # unrealized_pnl on Bybit api and unrealized_pnl on exchange UI
+    REALIZED_PNL = "realized_pnl"
     CLOSING_FEE = "closing_fee"
-    QUANTITY = "quantity"
-    SIZE = "size"
-    NOTIONAL = "notional"
+    CONTRACT_SIZE = "contract_size"
+    POSITION_SIZE = "position_size"
+    POSITION_VALUE = "position_value"
```
Herklos, line 424: Why renaming here?

octobot_trading/enums.py
```diff
@@ -407,6 +431,7 @@ class ExchangeConstantsPositionColumns(enum.Enum):
     MAINTENANCE_MARGIN_RATE = "maintenance_margin_rate"
     STATUS = "status"
     SIDE = "side"
+    ORIGINAL_SIDE = "original_side"
```
Herklos, line 434: As my first comment, I don't understand why we should keep the original side

octobot_trading/exchanges/connectors/exchange_simulator.py
```diff
@@ -208,7 +209,7 @@ def get_time_frames(self, importer):
                                                         set(self.exchange_manager.exchange_config.traded_time_frames)),
                                                    reverse=True)
 
-    def get_split_pair_from_exchange(self, pair) -> (str, str):
+    def get_split_pair_from_exchange(self, pair) -> typing.Tuple[str, str]:
```
Herklos, line 212: Why not keep as it?

octobot_trading/exchanges/implementations/margin_exchange_simulator.py
```diff
@@ -56,7 +57,7 @@ def get_exchange_current_time(self):
     def get_available_time_frames(self):
         return self.connector.get_available_time_frames()
 
-    def get_split_pair_from_exchange(self, pair) -> (str, str):
+    def get_split_pair_from_exchange(self, pair) -> typing.Tuple[str, str]:
```
Herklos, line 60: Why not keep as it?

octobot_trading/exchanges/parser/orders_parser.py
```diff
@@ -0,0 +1,730 @@
...
+    TraderOrderType,
+)
+from octobot_trading.enums import ExchangeConstantsMarketPropertyColumns as MarketCols
+from octobot_trading.enums import TradeOrderType as TradeOrderType
+import octobot_trading.exchanges.parser.util as parser_util
+
+
+class OrdersParser(parser_util.Parser):
+    """
+    use OrdersParser as a base class if you implement
```
Herklos, line 18: 👍

octobot_trading/exchanges/parser/orders_parser.py
```diff
@@ -0,0 +1,730 @@
...
+    a new parser for a non ccxt or crypto feed exchange
+
+    parser usage:   parser = OrdersParser(exchange)
+                    orders = await parser.parse_orders(raw_orders)
+                    order = await parser.parse_order(raw_order)
+    """
+
+    TEST_AND_FIX_SPOT_QUANTITIES: bool = False
+    TEST_AND_FIX_FUTURES_QUANTITIES: bool = False
```
Herklos, line 27: Why is it optional? (and disabled by default)

octobot_trading/exchanges/parser/orders_parser.py
```diff
@@ -0,0 +1,730 @@
...
+                    OrderCols.FILLED_AMOUNT.value,
+                    OrderCols.COST.value,
+                    OrderCols.REDUCE_ONLY.value,
+                    OrderCols.TAG.value,
+                    OrderCols.FEE.value,
+                ):
+                    self.set_fetched_attribute(fetched_order, key)
+
+    def set_fetched_attribute(self, fetched_order, key):
+        nothing = "nothing"
```
Herklos, line 708: ?

octobot_trading/exchanges/parser/orders_parser_ccxt.py
```diff
@@ -0,0 +1,66 @@
...
+    ID_KEYS: list = [CCXTOrderCols.ID.value]
+    SYMBOL_KEYS: list = [CCXTOrderCols.SYMBOL.value]
+    SIDE_KEYS: list = [CCXTOrderCols.SIDE.value]
+    TYPE_KEYS: list = [CCXTOrderCols.TYPE.value]
+    TAKER_OR_MAKER_KEYS: list = [CCXTOrderCols.TAKER_OR_MAKER.value]
+    PRICE_KEYS: list = [
+        CCXTOrderCols.AVERAGE.value,  # first try average as its more accurate
+        CCXTOrderCols.PRICE.value,
+        CCXTOrderCols.STOP_PRICE.value,  # use only if others are missing
+    ]
```
Herklos, line 32: I don't think it's a good approach, I think we should only use the standard one (the only one documented in https://docs.ccxt.com/en/latest/). We shouldn't brute force util we find a value. It may be a totally different value with some exchanges.

octobot_trading/exchanges/parser/orders_parser_generic_ccxt.py
```diff
@@ -0,0 +1,296 @@
...
+    """
+    Dont override this class, use CCXTOrdersParser instead
+
+    always/only include bulletproof custom code into the parser to improve generic support
+
+    parser usage:   parser = GenericCCXTOrdersParser(exchange)
+                    orders = await parser.parse_orders(raw_orders)
+                    order = await parser.parse_order(raw_order)
+    """
+
```
Herklos, line 25: I don't think we want a generic one. In my opinion we only want **one** ccxt parser that we can override to adapt.
Before officially supporting we have to do some specific tests, we can't just try it with generic parsers and say it's supported in my opinion.

octobot_trading/exchanges/parser/util.py
```diff
@@ -0,0 +1,424 @@
+import datetime
+import decimal
+from octobot_commons import timestamp_util
+from octobot_trading import constants as trading_constants
+from octobot_trading.enums import (
+    ExchangePositionCCXTColumns,
+)
+
+
+class Parser:
```
Herklos, line 10: I think the file name should be "parser.py"

octobot_trading/exchanges/parser/util.py
```diff
@@ -0,0 +1,424 @@
...
+                attribute_name,
+                error_messages_list,
+            ) in self.debugging_report_dict.items():
+                for index in range(len(error_messages_list)):
+                    if not error_messages_list[index] in error_messages_list[:index]:
+                        debugging_report_str += error_messages_list[index]
+                missing_attributes_str += f"{attribute_name}, "
+
+            self.exchange.logger.error(
+                self.debugging_report_template(
```
Herklos, line 58: I think this debugging report is a good idea but shouldn't managed only in parsers, it should be an octobot common thing

octobot_trading/exchanges/parser/util.py
```diff
@@ -0,0 +1,424 @@
...
+
+    def _log_missing_with_method(self, key_to_set, tried_message, method, error):
+        self._log_missing(
+            key_to_set=key_to_set,
+            tried_message=tried_message,
+            additional_message=f"with method: {type(method).__name__ if method else 'no method provided'} - error: {repr(error)}\n",
+        )
+
+
+def convert_any_time_to_seconds(raw_time):
```
Herklos, line 406: Why not two separate methods? Because your first check is a bit "random" and in most cases we know what is the exchange date format

octobot_trading/exchanges/parser/util.py
```diff
@@ -0,0 +1,424 @@
...
+            "^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n"
+            "This will help us to add support for your exchange\n"
+            "\n"
+            f"OctoBot is not able to parse all {self.PARSER_TITLE} attributes\n"
+            f"{report}"
+            "\n"
+            f"!!!!!!!!!!!!  DEBUGGING REPORT {self.PARSER_TITLE.upper()} PARSER END  !!!!!!!!!!!!"
+        )
+
+    def _try_to_find_and_set(
```
Herklos, line 89: I don't understand why we need so many "try to find" methods. CCXT and cryptofeed have documented standard keys, we can use them as default and when they're false or empty we do a specific implementation. I don't think it's a good idea to try to find by all means without really knowing what we used in the end.

octobot_trading/exchanges/parser/orders_parser_ccxt.py
techfreaque: oh I thought stop price is default, removed it.
No its optional for the generic parser

octobot_trading/exchanges/parser/orders_parser_generic_ccxt.py
techfreaque: Generic is not for the supported ones.
only the others you can select already, but aren't working well and aren't officially supported.
With the generic one, they have a way higher chance of success.

Also everything you think is not safe can be optional and turned off by default and later be an option in the coming experimental settings

But if you really don't want generic parser, it can be optional in the experimental settings, that will follow later

octobot_trading/exchanges/parser/util.py
techfreaque: I separated them and improved it, now its always going with standard first

octobot_trading/exchanges/parser/orders_parser.py
```diff
@@ -0,0 +1,730 @@
...
+                return price
+            if (
+                status == OrderStatus.CANCELED.value
+                or status == OrderStatus.OPEN.value
+                or status == OrderStatus.PENDING_CANCEL.value
+                or status == OrderStatus.EXPIRED.value
+                or status == OrderStatus.PENDING_CREATION.value
+                or status == OrderStatus.REJECTED.value
+            ):
+                return 0  # to check - should we set it to None?
```
GuillaumeDSM, line 438: price should be a decimal, I think it's safer to set it at 0 if we don't have it

octobot_trading/exchanges/parser/orders_parser.py
```diff
@@ -0,0 +1,730 @@
...
+
+    def missing_cost(self, _):
+        # check and should it be with fees?
+        filled_price = None
+        if (
+            filled_quantity := self.formatted_record.get(OrderCols.FILLED_AMOUNT.value)
+        ) is not None and (
+            filled_price := self.formatted_record.get(OrderCols.FILLED_PRICE.value)
+        ):
+            return filled_quantity * filled_price
```
GuillaumeDSM, line 470: I'm not sure we should do that as it's not always true. `filled_quantity * filled_price` is valid for spot but for futures it's different. Maybe we shouldn't have a fallback func here if we don't know in which case we are.

octobot_trading/exchanges/parser/orders_parser.py
```diff
@@ -0,0 +1,730 @@
...
+                or status == OrderStatus.CANCELED.value
+                or status == OrderStatus.REJECTED.value
+                or status == OrderStatus.EXPIRED.value
+                or status == OrderStatus.PENDING_CREATION.value
+                or status == OrderStatus.OPEN.value
+                or status == OrderStatus.PENDING_CANCEL.value
+            ):
+                return constants.ZERO
+            if status == OrderStatus.PARTIALLY_FILLED.value:
+                pass  # just here to let you know whats unhandled
```
GuillaumeDSM, line 558: can we add an error log here please ?

octobot_trading/exchanges/parser/orders_parser.py
```diff
@@ -0,0 +1,730 @@
...
+            return True
+        return None  # dont raise as it's optional
+
+    def _parse_tag(self):
+        # optional
+        self._try_to_find_and_set(
+            OrderCols.TAG.value,
+            self.TAG_KEYS,
+            use_info_sub_dict=self.USE_INFO_SUB_DICT_FOR_TAG,
+            allowed_falsely_values=(False,),
```
GuillaumeDSM, line 636: a tag can be `False` ?

octobot_trading/exchanges/parser/orders_parser_ccxt.py
```diff
@@ -0,0 +1,42 @@
+from octobot_trading.enums import (
+    ExchangeOrderCCXTColumns as CCXTOrderCols,
+)
+import octobot_trading.exchanges.parser.orders_parser as orders_parser
+
+
+class CCXTOrdersParser(orders_parser.OrdersParser):
```
GuillaumeDSM, line 7: 👍

octobot_trading/exchanges/parser/trades_parser.py
```diff
@@ -0,0 +1,61 @@
+from octobot_trading.enums import OrderStatus, TradeOrderType
+import octobot_trading.exchanges.parser as parser
+
+
+class TradesParser(parser.OrdersParser):
```
GuillaumeDSM, line 5: 👍

octobot_trading/personal_data/orders/order_util.py
```diff
@@ -370,15 +370,15 @@ async def get_order_size_portfolio_percent(exchange_manager, order_amount, side,
                                  portfolio_type=commons_constants.PORTFOLIO_TOTAL)
     if exchange_manager.is_future:
         # TODO check inverse
-        if market_quantity is constants.ZERO:
+        if market_quantity == constants.ZERO:
```
GuillaumeDSM, line 373: 👍

tests/personal_data/orders/test_order_util.py
```diff
@@ -457,33 +457,49 @@ async def test_ensure_orders_relevancy_with_positions(future_trader_simulator_wi
 @pytest.mark.asyncio
 async def test_get_order_size_portfolio_percent(trader_simulator):
     config, exchange_manager_inst, trader_inst = trader_simulator
-    api.force_set_mark_price(exchange_manager_inst, "BTC/UDST", 1000)
+    api.force_set_mark_price(exchange_manager_inst, "BTC/USDT", 1000)
```
GuillaumeDSM, line 460: 😂

octobot_trading/personal_data/orders/order.py
```diff
@@ -633,7 +633,7 @@ def to_dict(self):
             enums.ExchangeConstantsOrderColumns.AMOUNT.value: self.origin_quantity,
             enums.ExchangeConstantsOrderColumns.COST.value: self.total_cost,
             enums.ExchangeConstantsOrderColumns.QUANTITY_CURRENCY.value: self.quantity_currency,
-            enums.ExchangeConstantsOrderColumns.FILLED.value: self.filled_quantity,
+            enums.ExchangeConstantsOrderColumns.FILLED_AMOUNT.value: self.filled_quantity,
```
GuillaumeDSM, line 636: we will have to make sure everything using this will be updated before we release this

octobot_trading/personal_data/positions/position.py
```diff
@@ -734,16 +734,16 @@ def to_dict(self):
             enums.ExchangeConstantsPositionColumns.STATUS.value: self.status.value,
             enums.ExchangeConstantsPositionColumns.TIMESTAMP.value: self.timestamp,
             enums.ExchangeConstantsPositionColumns.SIDE.value: self.side.value,
-            enums.ExchangeConstantsPositionColumns.QUANTITY.value: self.quantity,
-            enums.ExchangeConstantsPositionColumns.SIZE.value: self.size,
-            enums.ExchangeConstantsPositionColumns.NOTIONAL.value: self.value,
+            enums.ExchangeConstantsPositionColumns.CONTRACT_SIZE.value: self.quantity,
+            enums.ExchangeConstantsPositionColumnsPOSITION_SIZE.value: self.size,
+            enums.ExchangeConstantsPositionColumns.POSITION_VALUE.value: self.value,
```
GuillaumeDSM, line 739: we will also have to ensure compatibiliy on those values for things using them

GuillaumeDSM (commented):
Thanks a lot for this all these improvements and making the PR easier to review @techfreaque !

octobot_trading/enums.py
techfreaque: yeah its there but not documented, same for the stopPrice
this is binance spot, bybit looks the same 
image

so I think we should add both stopPrice and fees to the default ccxt parser

octobot_trading/enums.py
techfreaque: I removed it thats for the positions parser

octobot_trading/exchanges/parser/util.py
techfreaque: oh yeah its changed

octobot_trading/exchanges/parser/util.py
techfreaque: the main point of try to find is:
1. check validity of the value
2. log wrong or missing values, so that we get a useful debugging report
3. having a standardized way to set values that fit for all - so we have less to maintain

If there is just one key and everything is alright, it won't do much except checking and setting the value

try to find is a bit misleading, maybe "check_and_set_value" would be better

"I don't think it's a good idea to try to find by all means without really knowing what we used in the end."
We dont try by all means, just what is defined in each parser implementation. 
If there is anything wrong, we get a precise report on what we used, what failed and why it failed

octobot_trading/personal_data/positions/position.py
techfreaque: I think I got all of them now (but better check as well), in octobot services is also a commit

octobot_trading/exchanges/connectors/exchange_simulator.py
techfreaque: black in VSCode lints it as an error and suggests this change, but with black in pycharm it doesnt care

octobot_trading/exchanges/parser/orders_parser.py
techfreaque: While we need it for bybit spot, there is also a high chance that it might cause issues with other exchanges.
As its using cost and remaining as a basis to recompute the other values

octobot_trading/exchanges/parser/util.py
techfreaque: I moved it to commons https://github.com/Drakkar-Software/OctoBot-Commons/pull/293

octobot_trading/exchanges/parser/parser.py
```diff
@@ -0,0 +1,452 @@
...
+            )
+        )
+
+
+def convert_timestamp_to_seconds(raw_timestamp: int or float or str) -> int:
+    try:
+        _time = int(raw_timestamp)
+    except:
+        raise ParserTimestampNotSecondsError
+    # change this before the year 5138
```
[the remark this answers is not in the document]
techfreaque: Future-proof. We should set a reminder as well

octobot_trading/exchanges/parser/orders_parser.py
techfreaque: okay, I removed the comments, its always decimal

octobot_trading/exchanges/parser/orders_parser.py
techfreaque: ooops copy/paste error

octobot_trading/exchanges/parser/orders_parser.py
techfreaque: oh right, i changed it, so it'll now only recalculate for spot.
Or do you think we should make it optional for spot?

octobot_trading/exchanges/parser/orders_parser.py
techfreaque: Actually that's not what we want, I removed it

octobot_trading/enums.py
Herklos: Sure, but if there is not "fee" anymore in ccxt response we may remove it. Our point is that we shouldn't have both "fees" and "fee"

octobot_trading/enums.py
Herklos: In my opinion it's not a parser role to do that. To me a parser a is mechanism to convert a format to another, not to implement or prepare internal logic.

octobot_trading/enums.py
Herklos: Is it supported by OrderParser?

octobot_trading/enums.py
Herklos: I don't understand why we need to add a "position" prefix to something in a PositionEnum

octobot_trading/exchanges/connectors/exchange_simulator.py
Herklos: Ok but it's not a good practice to update a code  that you have not modified. Can you revert it please?

octobot_trading/exchanges/implementations/margin_exchange_simulator.py
Herklos: Same here

octobot_trading/exchanges/parser/orders_parser_generic_ccxt.py
Herklos: Once again, it's a good idea but that's not the philosophy we have today. For now we want to add specific tentacle to support each exchange specifically. We want to be sure that we either fully support an exchange or not. We don't want to announce that we may support some part of an exchange with a generic algo which could produce unexpected effects if the exchange changes its logic.

octobot_trading/exchanges/parser/orders_parser_ccxt.py
```diff
@@ -0,0 +1,47 @@
...
+
+    TIMESTAMP_KEYS: list = [enums.ExchangeOrderCCXTColumns.TIMESTAMP.value]
+    STATUS_KEYS: list = [enums.ExchangeOrderCCXTColumns.STATUS.value]
+    ID_KEYS: list = [enums.ExchangeOrderCCXTColumns.ID.value]
+    SYMBOL_KEYS: list = [enums.ExchangeOrderCCXTColumns.SYMBOL.value]
+    SIDE_KEYS: list = [enums.ExchangeOrderCCXTColumns.SIDE.value]
+    TYPE_KEYS: list = [enums.ExchangeOrderCCXTColumns.TYPE.value]
+    TAKER_OR_MAKER_KEYS: list = [enums.ExchangeOrderCCXTColumns.TAKER_OR_MAKER.value]
+    PRICE_KEYS: list = [
+        enums.ExchangeOrderCCXTColumns.AVERAGE.value,  # first try average as its more accurate
```
Herklos, line 23: Is it documented somewhere in the ccxt documentation?

octobot_trading/exchanges/parser/orders_parser_ccxt.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+    override CCXTOrdersParser if you add support for a CCXT exchange
+    dont use GenericCCXTOrdersParser as a base
+
+    parser usage:   parser = CCXTOrdersParser(exchange)
+                    orders = await parser.parse_orders(raw_orders)
+                    order = await parser.parse_order(raw_order)
+    """
+
+    TIMESTAMP_KEYS: list = [enums.ExchangeOrderCCXTColumns.TIMESTAMP.value]
```
Herklos, line 15: I don't understand why we need those keys in the orders parser. I don't why we need to have it in the generic orders parsers but here we should only try ccxt well documented uniformized format in my opinion.

octobot_trading/exchanges/parser/orders_parser_ccxt.py
```diff
@@ -0,0 +1,47 @@
...
+    TYPE_KEYS: list = [enums.ExchangeOrderCCXTColumns.TYPE.value]
+    TAKER_OR_MAKER_KEYS: list = [enums.ExchangeOrderCCXTColumns.TAKER_OR_MAKER.value]
+    PRICE_KEYS: list = [
+        enums.ExchangeOrderCCXTColumns.AVERAGE.value,  # first try average as its more accurate
+        enums.ExchangeOrderCCXTColumns.PRICE.value,
+    ]
+    FILLED_PRICE_KEYS: list = [enums.ExchangeOrderCCXTColumns.AVERAGE.value]
+    AMOUNT_KEYS: list = [enums.ExchangeOrderCCXTColumns.AMOUNT.value]
+    REMAINING_KEYS: list = [enums.ExchangeOrderCCXTColumns.REMAINING.value]
+    FILLED_AMOUNT_KEYS: list = []  # todo
```
Herklos, line 29: Shouldn't it be the same as AMOUNT_KEYS?

octobot_trading/exchanges/parser/orders_parser_ccxt.py
```diff
@@ -0,0 +1,47 @@
...
+    REDUCE_ONLY_KEYS: list = [enums.ExchangeOrderCCXTColumns.REDUCE_ONLY.value]
+    POST_ONLY_KEYS: list = []
+    TAG_KEYS: list = [enums.ExchangeOrderCCXTColumns.TAG.value]
+    FEE_KEYS: list = [enums.ExchangeOrderCCXTColumns.FEE, enums.ExchangeOrderCCXTColumns.FEES.value]
+    QUANTITY_CURRENCY_KEYS: list = [enums.ExchangeOrderCCXTColumns.QUANTITY_CURRENCY.value]
+
+    def __init__(
+        self,
+        exchange,
+        parser_title: str = "ccxt orders",
```
Herklos, line 40: Shouldn't we use class attributes for naming?

octobot_trading/exchanges/parser/trades_parser.py
```diff
@@ -0,0 +1,69 @@
+import octobot_trading.enums as enums
+import octobot_trading.exchanges.parser as parser
+
+
+class TradesParser(parser.OrdersParser):
+    """
```
Herklos, line 6: I really like this parser, very clean👌

octobot_trading/exchanges/parser/trades_parser_ccxt.py
```diff
@@ -0,0 +1,24 @@
...
+
+        parser usage:   parser = TradesParser(exchange)
+                        trades = await parser.parse_trades(raw_trades)
+                        trade = await parser.parse_trade(raw_trade)
+    """
+
+    def __init__(
+        self,
+        exchange,
+        parser_title: str = "ccxt recent trades",
```
Herklos, line 17: Why "recent"?

octobot_trading/enums.py
techfreaque: Yeah the parser implementation is done. Only the Order class need to be change so its using the parsed value.
Depending on the exchange, it might require custom logic to work properly. I think binance should work out of the box

octobot_trading/enums.py
techfreaque: yeah you're right. I'll change the naming like that:
QUANTITY -> SINGLE_CONTRACT_VALUE
SIZE stays as it was
NOTIONAL -> VALUE

octobot_trading/enums.py
techfreaque: CCXT response is with both, that's why I added both, if you want only one, I'm not sure which one to keep

octobot_trading/exchanges/parser/orders_parser_generic_ccxt.py
techfreaque: what about the following:
The generic parser will never be used.
Except if it's enabled in the hidden experimental settings (which will follow later). Mainly for us to test existing fixes faster

octobot_trading/enums.py
```diff
@@ -103,6 +103,7 @@ class PositionStatus(enum.Enum):
     LIQUIDATED = "liquidated"
     OPEN = "open"
     ADL = "auto_deleveraging"
+    CLOSED = "closed"
```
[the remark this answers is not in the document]
techfreaque: there is no docs about that. status closed will only be set based on position size (size 0 means closed)

octobot_trading/enums.py
```diff
@@ -381,9 +387,22 @@ class ExchangeConstantsOrderColumns(enum.Enum):
     REDUCE_ONLY = "reduceOnly"
     STOP_PRICE = "stopPrice"
     TRIGGER_ABOVE = "triggerAbove"
+    POST_ONLY = "post_only"
     TAG = "tag"
 
 
+class CryptoFeedOrderColumns(enum.Enum):
```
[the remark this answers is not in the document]
techfreaque: cryptofeed order is an object
we convert it into a dict so that it works with the default orders parser (less to maintain)

for example:
order.status gets converted to order_dict[CryptoFeedOrderColumns.STATUS.value]
see cryptofeed_orders_parser 

or do you mean we should move the enum to cryptofeed parser?

octobot_trading/enums.py
```diff
@@ -381,9 +387,22 @@ class ExchangeConstantsOrderColumns(enum.Enum):
...
+    TIMESTAMP = "timestamp"
+    STATUS = "status"
+    ID = "id"
+    SYMBOL = "symbol"
```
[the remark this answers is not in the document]
techfreaque: it doesnt exist. see above

octobot_trading/exchanges/parser/orders_parser.py
```diff
@@ -0,0 +1,788 @@
...
+    REMAINING_KEYS: list = []
+    FILLED_AMOUNT_KEYS: list = []
+    COST_KEYS: list = []
+    REDUCE_ONLY_KEYS: list = []
+    POST_ONLY_KEYS: list = []
+    TAG_KEYS: list = []
+    FEE_KEYS: list = []
+    QUANTITY_CURRENCY_KEYS: list = []
+
+    USE_INFO_SUB_DICT_FOR_TIMESTAMP: bool = False
```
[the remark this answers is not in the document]
techfreaque: bybit requires it for reduce only and I'm sure there are many other cases

octobot_trading/exchanges/parser/orders_parser.py
```diff
@@ -0,0 +1,788 @@
...
+            not_found_method=self.missing_remaining,
+            use_info_sub_dict=self.USE_INFO_SUB_DICT_FOR_REMAINING,
+        )
+
+    def missing_remaining(self, _):
+        if status := self.formatted_record.get(
+            enums.ExchangeConstantsOrderColumns.STATUS.value
+        ):
+            if (
+                status == enums.OrderStatus.FILLED.value
```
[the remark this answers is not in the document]
techfreaque: Do you have an example? I just found that its only supported from 3.10
Or do you mean: 
If ...
elif ...
elif ...

octobot_trading/exchanges/parser/orders_parser.py
```diff
@@ -0,0 +1,788 @@
...
+        self._try_to_find_and_set(
+            enums.ExchangeConstantsOrderColumns.REDUCE_ONLY.value,
+            self.REDUCE_ONLY_KEYS,
+            not_found_method=self.missing_reduce_only,
+            use_info_sub_dict=self.USE_INFO_SUB_DICT_FOR_REDUCE_ONLY,
+            allowed_falsely_values=(False,),
+        )
+
+    def _parse_post_only(self):
+        # optional
```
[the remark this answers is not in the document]
techfreaque: as its not crucial and probably missing in many cases

octobot_trading/exchanges/parser/orders_parser_ccxt.py
techfreaque: They say average is the actual average filled price. So I would also say we prefer to use average when available.
As in current octobot master we also use this logic, I only moved the logic into the parser

octobot_trading/enums.py
techfreaque: I removed it

octobot_trading/exchanges/connectors/exchange_simulator.py
techfreaque: I changed it back

octobot_trading/enums.py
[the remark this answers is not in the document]
techfreaque: I removed it, its not necessary

octobot_trading/exchanges/parser/orders_parser_cryptofeed.py
```diff
@@ -0,0 +1,138 @@
...
+class CryptoFeedOrdersParser(orders_parser.OrdersParser):
+    """
+    override CryptoFeedOrdersParser if you add support for a crypto feed exchange
+
+    parser usage:   parser = CryptoFeedOrdersParser(exchange)
+                    orders = await parser.parse_orders(raw_orders)
+                    order = await parser.parse_order(raw_order)
+    """
+
+    TIMESTAMP_KEYS: list = [enums.CryptoFeedOrderColumns.TIMESTAMP.value]
```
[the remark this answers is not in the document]
techfreaque: I replaced the with the default ones.
We convert the cryptofeed order object into a dict first, so we can use the default orders parser (less to maintain)

octobot_trading/exchanges/parser/orders_parser_ccxt.py
techfreaque: Its actually ExchangeOrderCCXTColumns.FILLED, I changed it

octobot_trading/exchanges/parser/orders_parser_ccxt.py
techfreaque: yeah, I changed it

octobot_trading/exchanges/parser/trades_parser_ccxt.py
techfreaque: I changed it

octobot_trading/exchanges/parser/orders_parser.py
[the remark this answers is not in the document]
techfreaque: I changed it to status in ()

octobot_trading/enums.py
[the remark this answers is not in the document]
Herklos: As there is no documentation about this status (because it doesn't exist) I don't think we should use it.

octobot_trading/exchanges/connectors/exchange_simulator.py
```diff
@@ -14,6 +14,7 @@
 #  You should have received a copy of the GNU Lesser General Public
 #  License along with this library.
 import decimal
+import typing
```
Herklos, line 17: I think we can remove this import

octobot_trading/enums.py
Herklos: Can you check in the ccxt doc why there is two "fees" please?

octobot_trading/exchanges/parser/orders_parser.py
```diff
@@ -0,0 +1,753 @@
...
+        :param timestamp: to use if it's missing in the order
+
+        :param check_completeness: if true checks all attributes,
+            if somethings missing it'll try to fetch it from the exchange
+
+        :return: formatted order dict (100% complete or we raise NotImplemented)
+
+        """
+        check_completeness = (
+            check_completeness if check_completeness is not None else True
```
Herklos, line 226: Why check_completeness could be None?

octobot_trading/exchanges/parser/orders_parser.py
```diff
@@ -0,0 +1,753 @@
...
+            raise parser_util.ParserKeyNotFoundError
+
+    def _missing_octobot_order_type(self, missing_type_value):
+        taker_or_maker = None
+        if missing_type_value:
+            return enums.TraderOrderType(missing_type_value).value
+        if taker_or_maker := self.raw_record.get(
+            enums.ExchangeConstantsOrderColumns.TAKER_OR_MAKER.value
+        ):
+            # todo check - is it safe?
```
Herklos, line 357: ?

octobot_trading/exchanges/parser/orders_parser.py
```diff
@@ -0,0 +1,753 @@
...
+                return price
+            if status in (
+                enums.OrderStatus.CANCELED.value,
+                enums.OrderStatus.OPEN.value,
+                enums.OrderStatus.PENDING_CANCEL.value,
+                enums.OrderStatus.EXPIRED.value,
+                enums.OrderStatus.PENDING_CREATION.value,
+                enums.OrderStatus.REJECTED.value,
+            ):
+                return 0
```
Herklos, line 443: Is it the current (master) behavior?

octobot_trading/exchanges/parser/orders_parser.py
```diff
@@ -0,0 +1,753 @@
...
+            enums.ExchangeConstantsOrderColumns.STATUS.value
+        ):
+            if status in (
+                enums.OrderStatus.FILLED.value,
+                enums.OrderStatus.CLOSED.value,
+                enums.OrderStatus.CANCELED.value,
+                enums.OrderStatus.REJECTED.value,
+                enums.OrderStatus.EXPIRED.value,
+            ):
+                return 0
```
Herklos, line 535: Is it the current behavior?

techfreaque:
I'll close this PR and start a new one once I rebased my changes. As this PR history will not be relevant anymore.

PR #792, opened 2023-01-07
[CCXTWebsocket] migrate ws to CCXTWebsocketConnector

GuillaumeDSM:


octobot_trading/enums.py
```diff
@@ -407,6 +407,9 @@ class WebsocketFeeds(enum.Enum):
     MARK_PRICE = 'mark_price'
     LAST_PRICE = 'last_price'
     ORDERS = 'orders'
+    LEDGER = 'ledger'
```
Herklos, line 410: Is there a difference with PORTFOLIO ?

octobot_trading/exchanges/connectors/ccxt/ccxt_client_util.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+def get_symbols(client):
+    return set(client.symbols) \
+        if hasattr(client, "symbols") and client.symbols is not None else set()
+
+
+def get_time_frames(client):
+    return set(client.timeframes) \
+        if hasattr(client, "timeframes") and client.timeframes is not None else set()
```
Herklos, line 91: Should we use a try catch attribute error instead of checking with hasattr ?

octobot_trading/enums.py
GuillaumeDSM: in ccxt those are different feeds 
image

octobot_trading/exchanges/connectors/ccxt/ccxt_client_util.py
GuillaumeDSM: we would still need a if to check if it's None, I think we should leave it line this

Herklos: I don't understand why checking if value is None should be taken into account here.
I don't say using a try catch will simplify the code. My point is I think we should use hasattr only when it's necessary

octobot_trading/enums.py
Herklos: What's the difference?

GuillaumeDSM: image

octobot_trading/constants.py
```diff
@@ -57,6 +57,7 @@
     "ENABLE_EXCHANGE_HTTP_PROXY_FROM_ENV", "True")
 ENABLE_CCXT_VERBOSE = os_util.parse_boolean_environment_var("ENABLE_CCXT_VERBOSE", "False")
 ENABLE_CCXT_RATE_LIMIT = os_util.parse_boolean_environment_var("ENABLE_CCXT_RATE_LIMIT", "True")
+MIN_TIME_BETWEEN_PUBLIC_WS_UPDATES = float(os.getenv("MIN_TIME_BETWEEN_PUBLIC_WS_UPDATES", "0.1"))  # avoid spamming CPU
```
Herklos, line 60: Do we really need this?

Herklos: I mean, with that code, by default, we will drop many data. I'm not sure that's what we want when we implement websockets

octobot_trading/exchanges/abstract_websocket_exchange.py
```diff
@@ -113,8 +115,12 @@ async def subscribe(self):
         raise NotImplementedError("subscribe is not implemented")
 
     @abc.abstractmethod
-    async def close_and_restart_sockets(self, debounce_duration=0):
-        raise NotImplementedError("close_and_restart_sockets not implemented")
+    def update_followed_pairs(self):
+        raise NotImplementedError("_updated_followed_pairs not implemented")
```
Herklos, line 119:
```suggestion
        raise NotImplementedError("updated_followed_pairs not implemented")
```

octobot_trading/exchanges/connectors/ccxt/ccxt_client_util.py
```diff
@@ -0,0 +1,154 @@
...
+        return {'defaultType': 'future'}
+    if exchange_manager.is_margin:
+        return {'defaultType': 'margin'}
+    return {'defaultType': 'spot'}
+
+
+def get_symbols(client):
+    try:
+        return set(client.symbols)
+    except (AttributeError, TypeError):
```
Herklos, line 86: 👍

octobot_trading/exchanges/connectors/ccxt/ccxt_websocket_connector.py
```diff
@@ -13,64 +13,768 @@
...
 
     @classmethod
-    def get_ccxt_async_client(cls):
-        raise NotImplementedError("get_ccxt_async_client is not implemented")
+    def is_supporting_exchange(cls, exchange_candidate_name) -> bool:
+        return cls.get_name() == exchange_candidate_name
 
     def get_pair_from_exchange(self, pair):
+        """
+        Convert a cryptofeed symbol format to uniformized symbol format
```
Herklos, line 187:
```suggestion
        Convert a ccxt symbol format to uniformized symbol format
```

Herklos (commented):
Quick integration 👌 great work!

octobot_trading/enums.py
Herklos: Ok so should we disable account trades history when having ledger?

octobot_trading/exchanges/connectors/ccxt/ccxt_websocket_connector.py
```diff
@@ -466,20 +456,41 @@ def _subscribe_feed(self, feed, symbols=None, time_frame=None, since=None, limit
...
         self.logger.debug(f"Subscribed to {feed.value} {symbols_str}{time_frame_str}")
 
-    def _create_task_if_necessary(self, feed_callback, enable_timer_between_updates, feed_generator, **kwargs):
+    async def _feed_task(self, callback, enable_throttling, generator_func, *g_args, **g_kwargs):
+        while not self.should_stop:
+            try:
+                update_data = await generator_func(*g_args, **g_kwargs)
+                await callback(update_data, **g_kwargs)
+                if enable_throttling:
+                    # ccxt keeps updating the internal structures while waiting
```
Herklos, line 510: 👍

octobot_trading/exchanges/connectors/ccxt/ccxt_websocket_connector.py
```diff
@@ -39,6 +40,7 @@
 class CCXTWebsocketConnector(abstract_websocket.AbstractWebsocketExchange):
     INIT_REQUIRING_EXCHANGE_FEEDS = [Feeds.CANDLE]
     SUPPORTS_LIVE_PAIR_ADDITION = True
+    FEED_INITIALIZATION_TIMEOUT = 15 * commons_constants.MINUTE_TO_SECONDS
```
Herklos, line 43: So if we can't fetch the exchange candle histoiry we wait 15min right ? It's an edge case but I don't know if we should wait so much time

Herklos (approved):
💯

Herklos (approved):
:+1:

PR #796, opened 2023-01-15
WIP add RunAnalysis Mode and evaluators

techfreaque:


GuillaumeDSM:
is this a WIP ?

techfreaque:
> 

Yes its WIP, this is the part where the RunAnalysis factory needs to be implemented

GuillaumeDSM:
> > 
> 
> Yes its WIP, this is the part where the RunAnalysis factory needs to be implemented

Ok so here WIP doesn't mean that you are not done with the PR but rather that there is some work left to be done after merging ?

techfreaque:
> > > 
> > 
> > 
> > Yes its WIP, this is the part where the RunAnalysis factory needs to be implemented
> 
> Ok so here WIP doesn't mean that you are not done with the PR but rather that there is some work left to be done after merging ?

Its up to you if you wanna merge it before the RunAnalysis factory is done.
This PR is intended to be just a place to discuss the topic and share my first draft

techfreaque:
closed as this code is outdated

PR #803, opened 2023-01-15
add MinimumOrderCostNotReachedError and minor fixes

techfreaque:
MinimumOrderCostNotReachedError requires PR in tentacles and potentially some other places to properly handle the exception


octobot_trading/api/exchange.py
```diff
@@ -19,14 +19,15 @@
 
 import octobot_trading.constants
 import octobot_trading.enums
-import octobot_trading.exchanges.connectors.ccxt.enums
+import octobot_trading.exchanges.connectors.ccxt.enums as ccxt_enums
 import octobot_trading.exchanges as exchanges
+import octobot_trading.exchanges.exchange_builder as exchange_builder
```
Herklos, line 24: The import philosophy is to import high level module (exchanges, personal_data) when not importing from within the module. The idea here is to prevent changing all imports when we update internal module architecture

octobot_trading/errors.py
```diff
@@ -21,6 +21,12 @@ class MissingFunds(Exception):
     """
 
 
+class MinimumOrderCostNotReachedError(Exception):
```
Herklos, line 24: 👍

octobot_trading/exchanges/connectors/ccxt/ccxt_connector.py
```diff
@@ -497,13 +497,13 @@ def get_fees(self, symbol):
             return {
                 enums.ExchangeConstantsMarketPropertyColumns.TAKER.value:
                     market_status.get(enums.ExchangeConstantsMarketPropertyColumns.TAKER.value,
-                                      constants.CONFIG_DEFAULT_FEES),
+                                      ) or constants.CONFIG_DEFAULT_FEES,
```
Herklos, line 500: Why ?

octobot_trading/personal_data/orders/decimal_order_adapter.py
```diff
@@ -188,7 +188,11 @@ def decimal_check_and_adapt_order_details_if_necessary(quantity, price, symbol_m
 
             # check total_order_price not < min_cost
             if not personal_data.check_cost(float(total_order_price), min_cost):
-                return []
+                symbol = f"{symbol_market['symbol']} - " if 'symbol' in symbol_market else ""
+                raise errors.MinimumOrderCostNotReachedError(
```
Herklos, line 192: Then it should be catch if necessary everywhere this method is called

octobot_trading/personal_data/orders/order.py
```diff
@@ -714,7 +714,7 @@ def parse_order_type(raw_order):
                 parsed_order_type = _get_sell_and_buy_types(order_type)
         return side, parsed_order_type
     except (KeyError, ValueError):
-        return None, None
+        return side, enums.TraderOrderType.UNKNOWN
```
Herklos, line 717: Why ?

octobot_trading/exchanges/connectors/ccxt/ccxt_connector.py
techfreaque: When market_status[ExchangeConstantsMarketPropertyColumns.TAKER.value], market_status[ExchangeConstantsMarketPropertyColumns.MAKER.value] or market_status[ExchangeConstantsMarketPropertyColumns.FEE.value] is None, it will now properly use the fallback fees. Before my changes it would set them to None

octobot_trading/personal_data/orders/decimal_order_adapter.py
techfreaque: I added exceptions, also in the PR in octobot tentacles ->  [Order] handle MinimumOrderCostNotReachedError

octobot_trading/api/exchange.py
techfreaque: ah okay, I removed those changes

octobot_trading/personal_data/orders/order.py
techfreaque: Was an accident, I removed it

octobot_trading/errors.py
GuillaumeDSM: I don't really understand why we need both `MinimumOrderCostNotReachedError` and `MissingMinimalExchangeTradeVolume`. `MissingMinimalExchangeTradeVolume` is not doing the same thing ? Should we raise it where you are raising `MinimumOrderCostNotReachedError` or you need another exception ?

techfreaque: I changed it so its using MissingMinimalExchangeTradeVolume, didnt notice that one

GuillaumeDSM:
to reopen of relevant

PR #807, opened 2023-01-17
[Exchange] fix open position check and set_symbol_position_mode

techfreaque:


Herklos (approved):

octobot_trading/exchanges/traders/trader.pxd
```diff
@@ -40,4 +40,5 @@ cdef class Trader(util.Initializable):
     cpdef object set_risk(self, object risk)
     cpdef object convert_order_to_trade(self, object order)
 
-    cdef bint _has_open_position(self, str symbol)
```
Herklos, line 43: I don't remember if cpdef works with any()

techfreaque: It doenst work, I tried both cdef and cpdef

Herklos (commented):
Can you fix the tests please?

techfreaque:
> Can you fix the tests please?

Tests are fixed now

Herklos (approved):
Thanks for your fix!

octobot_trading/exchanges/connectors/ccxt/ccxt_connector.py
```diff
@@ -460,7 +460,7 @@ async def set_symbol_margin_type(self, symbol: str, isolated: bool):
                                                  marginType=self.CCXT_ISOLATED if isolated else self.CCXT_CROSSED)
 
     async def set_symbol_position_mode(self, symbol: str, one_way: bool):
-        return await self.client.set_position_mode(self, hedged=not one_way, symbol=symbol)
+        return await self.client.set_position_mode(hedged=not one_way, symbol=symbol)
```
GuillaumeDSM, line 463: 😮

octobot_trading/exchanges/traders/trader.pxd
GuillaumeDSM: generators don't work with cython. We gotter consumer the generator into a list and use this list into any() for it to be cythonizable. here it would mean doing this:
```
return any([
                position.size
                for position in self.exchange_manager.exchange_personal_data.positions_manager.get_symbol_positions(
                    symbol=symbol
                )])
```
Not sure it's worth it though. Usually doing this is faster and cythonizable:
```
for position in self.exchange_manager.exchange_personal_data.positions_manager.get_symbol_positions(
            symbol=symbol
    ):
    if position.size:
        return True
return False
```

octobot_trading/exchanges/traders/trader.py
```diff
@@ -671,5 +671,8 @@ def _has_open_position(self, symbol):
         :param symbol: the position symbol
         :return: True if open position for :symbol: exists
         """
-        return len(self.exchange_manager.exchange_personal_data.positions_manager.get_symbol_positions(
-            symbol=symbol)) != 0
+        return any(
```
GuillaumeDSM, line 674: wouldn't this change break the `set_position_mode` `if self._has_open_position(symbol):` condition and make it impossible to set a position mode on a empty position ? If yes then I think we should adapt `set_position_mode` to work on empty positions as well (I believe we should be able to change the position mode before creating the position)

techfreaque: I'm not sure if I get that right. As far as it looks to me, we can change the mode on an 0 size position, but we cant when size != 0.
I just tried it and it raises TooManyOpenPositionError when open position on the exchange and it doesnt when size is 0

octobot_trading/exchanges/traders/trader.pxd
techfreaque: I changed it to your second suggestion

octobot_trading/exchanges/traders/trader.py
GuillaumeDSM: Ok then let's forget what i said, your change is right 💯

octobot_trading/exchanges/traders/trader.pxd
GuillaumeDSM: Do you want to appl those changes or should we merge like this ?

PR #808, opened 2023-01-17
[BinanceUsdM] add to default futures exchanges

techfreaque:
This depends on [BinanceUsdM] PR in octobot-tentacles

### DIFF
--- a/octobot_trading/enums.py
+++ b/octobot_trading/enums.py
@@ -36,8 +36,14 @@ class PositionSide(enum.Enum):
 class TradeOrderType(enum.Enum):
     LIMIT = "limit"
     MARKET = "market"
+    # stop loss == reduce only on futures
     STOP_LOSS = "stop_loss"
     STOP_LOSS_LIMIT = "stop_loss_limit"
+    # conditional orders
+    # TODO handle them
+    CONDITIONAL_MARKET = "stop_market"
+    CONDITIONAL_LIMIT = "stop_limit"
+    
     TAKE_PROFIT = "take_profit"
     TAKE_PROFIT_LIMIT = "take_profit_limit"
     TRAILING_STOP = "trailing_stop"
--- a/octobot_trading/exchanges/connectors/ccxt/ccxt_connector.py
+++ b/octobot_trading/exchanges/connectors/ccxt/ccxt_connector.py
@@ -373,9 +373,46 @@ class CCXTConnector(abstract_exchange.AbstractExchange):
         )
 
     async def create_market_stop_loss_order(self, symbol, quantity, price, side, current_price, params=None) -> dict:
+        if self.client.has.get("createStopMarketOrder"):
+            try:
+                return self.adapter.adapt_order(
+                    await self.client.createStopMarketOrder(
+                        symbol,
+                        side=side,
+                        amount=quantity,
+                        stopPrice=price,
+                        params=params,
+                    )
+                )
+            except ccxt.OrderImmediatelyFillable:
+                # make sure stop always stops
+                created_order = await self.exchange_manager.exchange.create_order(
+                    order_type=(enums.TraderOrderType.BUY_MARKET
+                                if side == enums.TradeOrderSide.BUY.value
+                                else enums.TraderOrderType.SELL_MARKET),
+                    symbol=symbol,
+                    quantity=decimal.Decimal(str(quantity)),
+                    price=decimal.Decimal(str(price)),
+                    current_price=decimal.Decimal(str(current_price))
+                )
+                created_order[enums.ExchangeConstantsOrderColumns.TYPE.value] = (
+                    enums.TraderOrderType.STOP_LOSS.value
+                    )
+                return created_order
         raise NotImplementedError("create_market_stop_loss_order is not implemented")
 
-    async def create_limit_stop_loss_order(self, symbol, quantity, price=None, side=None, params=None) -> dict:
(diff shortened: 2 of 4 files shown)


octobot_trading/enums.py
```diff
@@ -31,8 +31,14 @@ class PositionSide(enum.Enum):
 class TradeOrderType(enum.Enum):
     LIMIT = "limit"
     MARKET = "market"
+    # stop loss == reduce only on futures
     STOP_LOSS = "stop_loss"
     STOP_LOSS_LIMIT = "stop_loss_limit"
+    # conditional orders
+    # TODO handle them
+    STOP_MARKET = "stop_market"
+    STOP_LIMIT = "stop_limit"
```
Herklos, line 40: I don't understand the difference between STOP_LIMIT and STOP_LOSS_LIMIT

techfreaque: So those order types are only relevant for futures
STOP_LIMIT can be an order that opens or closes a position, when set to reduce only a STOP_LIMIT will be the same as an STOP_LOSS_LIMIT

So the STOP_LOSS_LIMIT will only close positions, while STOP_LIMIT can open or close positions

techfreaque: same applies for STOP_MARKET / STOP_LOSS_MARKET

GuillaumeDSM: should we name them CONDITIONAL_MARKET and CONDITIONAL_LIMIT ? That might be less confusing

octobot_trading/exchanges/connectors/ccxt/ccxt_connector.py
```diff
@@ -358,9 +359,46 @@ async def create_limit_sell_order(self, symbol, quantity, price=None, params=Non
...
+                return self.adapter.adapt_order(
+                    await self.client.createStopMarketOrder(
+                        symbol,
+                        side=side,
+                        amount=quantity,
+                        stopPrice=price,
+                        params=params,
+                    )
+                )
+            except OrderImmediatelyFillable:
```
GuillaumeDSM, line 373:
```suggestion
            except ccxt.OrderImmediatelyFillable:
```

octobot_trading/exchanges/connectors/ccxt/ccxt_connector.py
```diff
@@ -18,6 +18,7 @@
 import decimal
 
 import ccxt.async_support as ccxt
+from ccxt.async_support import OrderImmediatelyFillable
```
GuillaumeDSM, line 21: I don't think we need this import

octobot_trading/enums.py
techfreaque: I renamed it to conditional

octobot_trading/exchanges/connectors/ccxt/ccxt_connector.py
techfreaque: done

octobot_trading/exchanges/connectors/ccxt/ccxt_connector.py
techfreaque: Ots removed

GuillaumeDSM (approved):
good for me !

Herklos (approved):

GuillaumeDSM:
When the linter is fixed, we can merge it

techfreaque:
> When the linter is fixed, we can merge it

Should be good now

GuillaumeDSM:
We can merge it as soon as the CI passes

octobot_trading/exchanges/types/rest_exchange.py
```diff
@@ -264,7 +268,7 @@ async def _create_specific_order(self, order_type, symbol, quantity: decimal.Dec
                                                                       params=params)
         elif order_type == enums.TraderOrderType.STOP_LOSS_LIMIT:
             created_order = await self._create_limit_stop_loss_order(symbol, float_quantity, price=float_price,
-                                                                     side=side, params=params)
+                                                                     side=side, stop_price=stop_price, params=params)
```
Herklos, line 271: I don't understand why we need a new param here. Why not using price?

octobot_trading/exchanges/connectors/ccxt/ccxt_connector.py
```diff
@@ -373,9 +373,46 @@ async def create_limit_sell_order(self, symbol, quantity, price=None, params=Non
...
+                return self.adapter.adapt_order(
+                    await self.client.createStopMarketOrder(
+                        symbol,
+                        side=side,
+                        amount=quantity,
+                        stopPrice=price,
+                        params=params,
+                    )
+                )
+            except ccxt.OrderImmediatelyFillable:
```
Herklos, line 387: Is this behavior documented somewhere in the ccxt documentation?

techfreaque: Havent found it documented anywhere, but if you run something like arbitrage mode, this could get raised when you have volatility. You were actually ending up with no stop loss order getting placed or triggered

octobot_trading/exchanges/types/rest_exchange.py
techfreaque: stop limit orders have two parameter for price.
price will be the actual limit order price of the limit order.
The limit order gets created by the exchange, when the mark price crosses stop_price.

Herklos: Right :+1: Thanks!

Herklos (approved):
Thanks @techfreaque!

PR #815, opened 2023-01-28
[Trade] properly handle status expired 

techfreaque:
If the trade status is expired it now handle it properly

### DIFF
--- a/octobot_trading/personal_data/trades/trade.py
+++ b/octobot_trading/personal_data/trades/trade.py
@@ -84,10 +84,16 @@ class Trade:
         self.tag = order.tag
 
     def get_time(self):
-        return self.executed_time if self.status is not enums.OrderStatus.CANCELED else self.canceled_time
+        return (self.executed_time
+                if (self.status is not enums.OrderStatus.CANCELED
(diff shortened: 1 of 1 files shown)


octobot_trading/personal_data/trades/trade.py
```diff
@@ -84,10 +84,16 @@ def update_from_order(self, order, creation_time=0, canceled_time=0, executed_ti
         self.tag = order.tag
 
     def get_time(self):
-        return self.executed_time if self.status is not enums.OrderStatus.CANCELED else self.canceled_time
+        return (self.executed_time
+                if (self.status is not enums.OrderStatus.CANCELED
+                    and self.status is not enums.OrderStatus.EXPIRED)
+                else self.canceled_time)
```
Herklos, line 90: Shouldn't it be expired time if EXPIRED?

techfreaque: could be, but as an expired order is just a canceled order that got automatically canceled by the exchange. I'd say we are good like that

Herklos (approved):

GuillaumeDSM (approved):

PR #821, opened 2023-02-02
[Ticker] properly handle exchanges with MARK_PRICE_IN_TICKER

techfreaque:
Binance usdm uses MARK_PRICE_IN_TICKER which is getting treated as a secondary source and mark price will get overwritten by mark price from recent trades.

With this change mark price from ticker will work properly

This was not used before by, so thats why nobody noticed

### DIFF
--- a/octobot_trading/exchange_data/ticker/channel/ticker.py
+++ b/octobot_trading/exchange_data/ticker/channel/ticker.py
@@ -103,7 +103,7 @@ class TickerProducer(exchanges_channel.ExchangeChannelProducer):
                                              self.channel.exchange_manager.id).get_internal_producer().push(
                 symbol,
                 decimal.Decimal(str(ticker[enums.ExchangeConstantsMarkPriceColumns.MARK_PRICE.value])),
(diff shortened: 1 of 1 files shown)


GuillaumeDSM (approved):
👍

GuillaumeDSM:
@Herklos I let you merge it when you approve it !

Herklos:
We should still handle TICKER_CLOSE_PRICE somewhere. I don't think that just because Binance provides it in the ticker, it should be generalized for all exchanges.

techfreaque:
> We should still handle TICKER_CLOSE_PRICE somewhere. I don't think that just because Binance provides it in the ticker, it should be generalized for all exchanges.

This code is/was never used by any other exchange. As you must set MARK_PRICE_IN_TICKER to use it. This is only relevant to futures and doesnt have anything to do with TICKER_CLOSE_PRICE.
TICKER_CLOSE_PRICE was wrongly used here, it should be EXCHANGE_MARK_PRICE as its for cases where the actual real mark price (not the close) is in the ticker

GuillaumeDSM:
to reopen of relevant

PR #846, opened 2023-02-21
[Orders] handle failed cancel

GuillaumeDSM:


octobot_trading/exchanges/traders/trader.py
```diff
@@ -326,46 +326,86 @@ async def cancel_order(self, order: object, ignored_order: object = None,
...
+            raise errors.OpenOrderError(
+                "Order is open, but can't be cancelled. This is unexpected"
+            ) from err
+        elif order.is_filled():
+            raise errors.FilledOrderError("Order is filled, it can't be cancelled") from err
+        elif order.is_closed():
+            raise errors.ClosedOrderError("Order is closed, it can't be cancelled") from err
+        else:
+            # should not happen
+            raise errors.OrderCancelError(
```
Herklos, line 400: Can this raise produce unexpected effects? In other words, do we need to catch this error somewhere?

octobot_trading/exchanges/traders/trader.py
```diff
@@ -326,46 +326,86 @@ async def cancel_order(self, order: object, ignored_order: object = None,
...
+            await order.state.synchronize(force_synchronization=True)
+            if previous_status != order.status:
+                # status changed: wait for state change
+                await order.state.wait_for_next_state(cancelling_timeout)
+        if order.is_cancelled():
+            self.logger.debug(f"Tried to cancel an already cancelled order.")
+            return True
+        if order.is_cancelling():
+            if wait_for_cancelling:
+                await self._wait_for_order_cancel(order, cancelling_timeout)
```
Herklos, line 388: 👍

octobot_trading/exchanges/traders/trader.py
GuillaumeDSM: this is was happens at the moment (before PR), it will forward the error to the caller. I think we should leave it like this as it's a real problem if it happens

Herklos: I agree. The caller is the order updater right?

GuillaumeDSM: Here the caller is the thing that called trader.cancel_order(), usually a trading mode or a user command

Herklos: Ok. Today the error is forwarded to the caller but don't you think we should add a catch in trading modes to prevent unexpected behaviors?

GuillaumeDSM: It's the same as for other unexpected state errors (when the order is filled etc), I think it's more on a trading mode by trading mode basis to decide. I'm not sure we can have a generic thing here. What is your opinion ?

Herklos: I agree for specific calls but for more generic ones like the cancel all at startup I think we should add something

octobot_trading/exchanges/traders/trader.py
```diff
@@ -368,9 +368,19 @@ async def _handle_order_cancellation(self, order, ignored_order,
         return True
 
     async def _handle_order_cancel_error(self, order, err, wait_for_cancelling, cancelling_timeout):
-        # can't cancel order: it likely means that the order is not open on exchange anymore
+        """
+        Use when an order can't be cancelled: it usually means that the order is not open on exchange anymore.
+        Will synch the given order on exchange to figure out the cancel issue if any.
+        Returns True when the order cancel ends up successful even though an error initially occurred
+        Raises OrderCancelError on unrecoverable cases
+        Raises a subclass of UnexpectedExchangeSideOrderStateError when the order is in a unexpected state
```
Herklos, line 376: 👍

Herklos (approved):
👍

PR #861, opened 2023-03-10
[Keywords] add QUOTE_VALUE QuantityType

techfreaque:
adds the syntax 100$ to define the amount based on the value in quote currency.
On a BTC/EUR pair the syntax would still be 100$ to get a 100 EUR position size, so we might want a different syntax.

### DIFF
--- a/octobot_trading/modes/script_keywords/basic_keywords/amount.py
+++ b/octobot_trading/modes/script_keywords/basic_keywords/amount.py
@@ -42,6 +42,16 @@ async def get_amount_from_input_amount(
     elif amount_type is dsl.QuantityType.AVAILABLE_PERCENT:
         amount_value = await account_balance.available_account_balance(context, side, reduce_only=reduce_only) \
                        * amount_value / 100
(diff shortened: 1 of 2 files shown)


techfreaque:
Not sure why CI is failing on the import inside the method. (If I import it at the top of the file I'm getting a circular import error)

octobot_trading/modes/script_keywords/basic_keywords/amount.py
```diff
@@ -42,6 +42,16 @@ async def get_amount_from_input_amount(
...
                        * amount_value / 100
+    elif amount_type is dsl.QuantityType.QUOTE_VALUE:
+        if (context.exchange_manager.is_future 
+            and context.exchange_manager.exchange.is_inverse_symbol(context.symbol)
+            ):
+            # is measured in QUOTE_VALUE
+            pass
+        else:
+            # avoid circular import error
+            import tentacles.Meta.Keywords.scripting_library.data.reading.exchange_public_data as exchange_public_data
```
Herklos, line 53: Is there another way to handle imports in the scripting library to avoid circular imports?

PR #862, opened 2023-03-11
[OrdersStorage] Init orders storage

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Commons/pull/335


octobot_trading/personal_data/orders/groups/group_util.py
```diff
@@ -0,0 +1,24 @@
...
+#  MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+#  Lesser General Public License for more details.
+#
+#  You should have received a copy of the GNU Lesser General Public
+#  License along with this library.
+import octobot_trading.personal_data.orders.order_group as order_group
+import octobot_commons.tentacles_management as tentacles_management
+
+
+def get_group_type(group_type_str: str):
```
Herklos, line 20:
```suggestion
def get_group_type_class(group_type_str: str):
```

octobot_trading/enums.py
```diff
@@ -526,3 +527,15 @@ class TradingSignalOrdersActions(enum.Enum):
     ADD_TO_GROUP = "add_to_group"
     EDIT = "edit"
     CANCEL = "cancel"
+
+
+class PersistedOrdersAttr(enum.Enum):
```
Herklos, line 532: I would prefer
```suggestion
class StoredOrdersAttr(enum.Enum):
```

octobot_trading/personal_data/orders/channel/orders.py
```diff
@@ -141,6 +159,38 @@ async def handle_post_open_orders_update(
             )
             await self._complete_open_order_init(waiting_complete_init_orders, is_from_bot)
 
+    async def _complete_pending_groups_from_storage(self, pending_groups):
+        # create order groups and their associated self-managed orders if any
+        # use copy.copy(pending_groups) as new groups might be added from new orders,
+        # in this case, also create the associated new orders
+        to_complete_groups = list(pending_groups.keys())
+        completed_groups = set()
+        max_allowed_nested_chained_orders_groups = 100
```
Herklos, line 168: I think we should create a constant somewhere to store this value

octobot_trading/personal_data/orders/channel/orders.py
```diff
@@ -141,6 +159,38 @@ async def handle_post_open_orders_update(
...
+        for _ in range(max_allowed_nested_chained_orders_groups):
```
Herklos, line 169: Why do we need it?

octobot_trading/personal_data/orders/channel/orders.py
```diff
@@ -141,6 +159,38 @@ async def handle_post_open_orders_update(
             )
             await self._complete_open_order_init(waiting_complete_init_orders, is_from_bot)
 
+    async def _complete_pending_groups_from_storage(self, pending_groups):
+        # create order groups and their associated self-managed orders if any
+        # use copy.copy(pending_groups) as new groups might be added from new orders,
+        # in this case, also create the associated new orders
+        to_complete_groups = list(pending_groups.keys())
```
Herklos, line 166: How `pending_groups` is filled?

octobot_trading/personal_data/orders/order_util.py
```diff
@@ -435,3 +437,55 @@ async def get_order_size_portfolio_percent(exchange_manager, order_amount, side,
...
+            chained_order_inst = await create_order_from_order_storage_details(
+                chained_order, exchange_manager, pending_groups
+            )
+            await chained_order_inst.set_as_chained_order(
+                order,
+                chained_order[enums.PersistedOrdersAttr.HAS_BEEN_BUNDLED.value],
+                chained_order[enums.PersistedOrdersAttr.EXCHANGE_CREATION_PARAMS.value],
+                **chained_order[enums.PersistedOrdersAttr.TRADER_CREATION_KWARGS.value],
+            )
+            order.add_chained_order(chained_order_inst)
```
Herklos, line 491: I think we should put this code somewhere else, I was surprised to find an order creation in a method named `_update_from_order_details`. I think we need a method that updates order from storage data and a second method that creates chained orders if a loaded order from storage had chained orders.

octobot_trading/personal_data/orders/order_util.py
```diff
@@ -435,3 +437,55 @@ async def get_order_size_portfolio_percent(exchange_manager, order_amount, side,
...
+    order.associated_entry_ids = order_details[enums.PersistedOrdersAttr.ENTRIES.value]
+    group = order_details[enums.PersistedOrdersAttr.GROUP.value]
+    if group:
+        group_name = group[enums.PersistedOrdersAttr.GROUP_ID.value]
+        if group_name not in pending_groups:
+            pending_groups[group_name] = exchange_manager.exchange_personal_data.orders_manager.get_or_create_group(
+                group_util.get_group_type(group[enums.PersistedOrdersAttr.GROUP_TYPE.value]),
+                group_name,
+            )
+        order.order_group = pending_groups[group_name]
```
Herklos, line 478: I think we should create a new method that adds an order to a group

octobot_trading/personal_data/orders/order_util.py
```diff
@@ -435,3 +437,55 @@ async def get_order_size_portfolio_percent(exchange_manager, order_amount, side,
...
+
+async def _update_from_order_details(order, order_details, exchange_manager, pending_groups):
+    # rebind order attributes that are not stored on exchange
+    order_dict = order_details[orders_storage.OrdersStorage.ORIGIN_VALUE_KEY]
+    order.tag = order_dict[enums.ExchangeConstantsOrderColumns.TAG.value]
+    order.trader_creation_kwargs = order_details[enums.PersistedOrdersAttr.TRADER_CREATION_KWARGS.value]
+    order.exchange_creation_params = order_details[enums.PersistedOrdersAttr.EXCHANGE_CREATION_PARAMS.value]
+    order.set_shared_signal_order_id(order_details[enums.PersistedOrdersAttr.SHARED_SIGNAL_ORDER_ID.value])
+    order.has_been_bundled = order_details[enums.PersistedOrdersAttr.HAS_BEEN_BUNDLED.value]
+    order.associated_entry_ids = order_details[enums.PersistedOrdersAttr.ENTRIES.value]
```
Herklos, line 469: Shouldn't we add default values if the key doesn't exist?

octobot_trading/personal_data/orders/order_util.py
```diff
@@ -435,3 +437,55 @@ async def get_order_size_portfolio_percent(exchange_manager, order_amount, side,
...
+
+
+async def update_from_order_storage(order, exchange_manager, pending_groups):
+    # only real orders can be updated by stored orders
+    if exchange_manager.is_trader_simulated or exchange_manager.is_backtesting \
+            or not exchange_manager.storage_manager.orders_storage:
+        return
+    order_details = await exchange_manager.storage_manager.orders_storage.get_startup_order_details(order.order_id)
+    if order_details:
+        await _update_from_order_details(order, order_details, exchange_manager, pending_groups)
```
Herklos, line 449: Should we return here?

tests/personal_data/orders/test_order_util.py
```diff
@@ -488,3 +501,135 @@ async def test_get_order_size_portfolio_percent(trader_simulator):
...
+        order_storage_details, exchange_manager_inst, pending_groups
+    )
+    assert pending_groups == {
+        group_1.name: group_1,
+        group_2.name: group_2,
+    }
+    chained_orders = created_order.chained_orders
+    assert len(chained_orders) == 2
+    for chained_order in chained_orders:
+        assert chained_order.order_group is group_1
```
Herklos, line 629: 👍

octobot_trading/personal_data/orders/order_util.py
```diff
@@ -435,3 +437,55 @@ async def get_order_size_portfolio_percent(exchange_manager, order_amount, side,
...
+
+
+async def create_order_from_order_storage_details(order_storage_details, exchange_manager, pending_groups):
```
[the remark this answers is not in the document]
GuillaumeDSM: right, good idea

octobot_trading/personal_data/orders/groups/group_util.py
GuillaumeDSM: I'm chaning it to `get_group_class`

octobot_trading/personal_data/orders/channel/orders.py
GuillaumeDSM: I don't think so. I put it under a variable to add meaning instead of a magic number but it's something that will almost never happen, I don't think we should polute constants with it

octobot_trading/personal_data/orders/channel/orders.py
GuillaumeDSM: we do :/ it will probably almost never be useful but we need to handle the case as it is possible to loop indefinitely otherwise

octobot_trading/personal_data/orders/order_util.py
GuillaumeDSM: All those keys are always filled in format_order. To me adding default values is not required

octobot_trading/personal_data/orders/order_util.py
GuillaumeDSM: the is no return to the function, i'm not sure i understand what you mean

octobot_trading/personal_data/orders/groups/group_util.py
Herklos: 👍

octobot_trading/personal_data/orders/order_util.py
Herklos: So we accept that this code can raise a key error which prevents the order from been loaded ? It could be for any reason like a format change

GuillaumeDSM: you are right, i'll make it so that it doest crash if the format changes

octobot_trading/personal_data/orders/order.py
```diff
@@ -25,8 +25,11 @@
 import octobot_trading.enums as enums
 import octobot_trading.errors as errors
 import octobot_trading.personal_data.orders.states as orders_states
+import octobot_trading.personal_data.orders.order_factory as order_factory
```
Herklos, line 28: This is the kind of import that makes me think there's an architecture issue here

octobot_trading/personal_data/orders/order.py
```diff
@@ -604,6 +609,56 @@ async def update_from_order(self, other_order):
...
+                chained_order_inst = await order_factory.create_order_from_order_storage_details(
+                    chained_order, exchange_manager, pending_groups
+                )
+                await chained_order_inst.set_as_chained_order(
+                    self,
+                    chained_order.get(enums.StoredOrdersAttr.HAS_BEEN_BUNDLED.value, False),
+                    chained_order.get(enums.StoredOrdersAttr.EXCHANGE_CREATION_PARAMS.value, {}),
+                    **chained_order.get(enums.StoredOrdersAttr.TRADER_CREATION_KWARGS.value, {}),
+                )
+                self.add_chained_order(chained_order_inst)
```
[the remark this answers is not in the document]
GuillaumeDSM: I'm moving _restore_chained_orders_from_storage_order_details to order_factory, this will also remove the import in order.py.
I'll move _restore_order_group_from_storage_order_details to group_factory

octobot_trading/personal_data/orders/groups/group_util.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def get_group_class(group_type_str: str):
+    for group_class in tentacles_management.get_all_classes_from_parent(order_group.OrderGroup):
+        if group_type_str == group_class.__name__:
+            return group_class
+    raise KeyError(group_type_str)
+
+
+def get_or_create_order_group_from_storage_order_details(order_details, exchange_manager):
```
Herklos, line 29: 👍

octobot_trading/personal_data/orders/order_util.py
```diff
@@ -435,3 +437,68 @@ async def get_order_size_portfolio_percent(exchange_manager, order_amount, side,
...
+    if not exchange_manager.storage_manager.orders_storage \
+            or not exchange_manager.storage_manager.orders_storage.should_store_date():
+        return
+    order_details = await exchange_manager.storage_manager.orders_storage.get_startup_order_details(order.order_id)
+    if order_details:
+        order.update_from_storage_order_details(order_details)
+        await create_orders_storage_related_elements(order, order_details, exchange_manager, pending_groups)
+
+
+async def create_orders_storage_related_elements(order, order_storage_details, exchange_manager, pending_groups):
```
Herklos, line 453: I think we should also put it in order factory

Herklos (approved):

octobot_trading/personal_data/orders/order_util.py
GuillaumeDSM: I didn't put it there because it's also creating groups, but we could also put it there. `_create_storage_self_managed_orders_from_group` could also be but there as this one it is creating orders

Herklos (approved):
Great feature! 💯

PR #866, opened 2023-03-15
[RestExchange] filter symbols depending on exchange type

techfreaque:
Makes sure a spot exchange will only run on a spot pair and a futures pair only on futures


octobot_trading/exchanges/types/rest_exchange.py
```diff
@@ -508,6 +509,17 @@ def get_split_pair_from_exchange(self, pair) -> (str, str):
...
         return self.connector.get_exchange_pair(pair)
 
+    def get_exchange_pairs(self) -> set:
+        # make sure spot doesnt use futures pairs and vice versa
+        if self.exchange_manager.is_spot_only or self.exchange_manager.is_margin: 
+            return {symbol for symbol in self.connector.symbols 
+                    if octobot_commons.SETTLEMENT_ASSET_SEPARATOR not in symbol}
+        elif self.exchange_manager.is_future:
+            return {symbol for symbol in self.connector.symbols 
+                    if octobot_commons.SETTLEMENT_ASSET_SEPARATOR in symbol}
```
GuillaumeDSM, line 519: Good idea !
I think it would be better to use `parse_symbol(symbol).is_future()` and `parse_symbol(symbol).is_spot()`

techfreaque: oh yeah looks better, I changed it

Herklos (approved):

PR #876, opened 2023-03-23
[CcxtAdapter] fix default parse_funding_rate

techfreaque:
Not sure if I misunderstand something, but I think it should be like this.


techfreaque:
required for https://github.com/Drakkar-Software/OctoBot-Tentacles/pull/794

GuillaumeDSM:
the idea with funding is that we keep the previous funding rate and the next one. According to my experience with ccxt, PREVIOUS_FUNDING_RATE is the last one and FUNDING_RATE is the next one. But that may change depending on exchanges. is it like this with binanceusdm ?

PR #896, opened 2023-04-20
[Orders] handle pending creation orders failed first fetch

GuillaumeDSM:


octobot_trading/personal_data/orders/order.py
```diff
@@ -385,7 +388,7 @@ async def on_open(self, force_open=False, is_from_exchange_data=False):
 
     async def on_fill(self, force_fill=False, is_from_exchange_data=False):
         logging.get_logger(self.get_logger_name()).debug(f"on_fill triggered for {self}")
-        if self.is_open() and not self.is_refreshing():
+        if (self.is_open() and not self.is_refreshing()) or self.is_pending_creation():
```
Herklos, line 391: Should we also check is_refreshing ?

GuillaumeDSM: in the particular case of `is_pending_creation`,  we are in the sync from exchange call of pending creation orders. that means that it's still refreshing but we should proceed anyway. In other words, we skip the open state of the order and we go directly to filled (from pending creation)

octobot_trading/personal_data/orders/states/pending_creation_order_state.py
```diff
@@ -23,6 +29,48 @@ def __init__(self, order, is_from_exchange_data):
...
+            while self.order.is_pending_creation() \
+                    and time.time() - t0 < octobot_trading.constants.INDIVIDUAL_ORDER_SYNC_TIMEOUT:
+                iteration += 1
+                await exchange_channel.get_chan(
+                    octobot_trading.constants.ORDERS_CHANNEL,
+                    self.order.exchange_manager.id
+                ).get_internal_producer().update_order_from_exchange(
+                    order=self.order,
+                    wait_for_refresh=True,
+                    force_job_execution=force_synchronization,
```
Herklos, line 56: Is it working if force_synchronization is False?

GuillaumeDSM: right, i'll change it to true

octobot_trading/constants.py
```diff
@@ -128,7 +128,7 @@
 BALANCE_CHANNEL = "Balance"
 BALANCE_PROFITABILITY_CHANNEL = "BalanceProfitability"
 POSITIONS_CHANNEL = "Positions"
-INDIVIDUAL_ORDER_SYNC_TIMEOUT = 3 * commons_constants.MINUTE_TO_SECONDS
+INDIVIDUAL_ORDER_SYNC_TIMEOUT = 1 * commons_constants.MINUTE_TO_SECONDS
```
Herklos, line 131: 👍

octobot_trading/exchanges/types/rest_exchange.py
```diff
@@ -54,6 +54,8 @@ class RestExchange(abstract_exchange.AbstractExchange):
     # set True when get_positions() is not returning empty positions and should use get_position() instead
     REQUIRES_SYMBOL_FOR_EMPTY_POSITION = False
     SUPPORTS_SET_MARGIN_TYPE = True  # set False when there is no API to switch between cross and isolated margin types
+    EXPECT_POSSIBLE_ORDER_NOT_FOUND_ON_INSTANTLY_FILLED_ORDERS = False  # set True when get_order() can return None
```
Herklos, line 57: 👍

Herklos (approved):

PR #898, opened 2023-04-21
Ccxt up

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Trading/pull/896

### DIFF
--- a/octobot_trading/exchanges/connectors/ccxt/ccxt_websocket_connector.py
+++ b/octobot_trading/exchanges/connectors/ccxt/ccxt_websocket_connector.py
@@ -632,6 +632,8 @@ class CCXTWebsocketConnector(abstract_websocket_exchange.AbstractWebsocketExchan
         is_initialized_func = None
         if feed is Feeds.CANDLE:
             def candle_is_initialized_func():
+                if self.exchange_manager is None:
+                    return False
                 try:
                     return self.exchange_manager.exchange_symbols_data.get_exchange_symbol_data(
                         g_kwargs["symbol"], allow_creation=False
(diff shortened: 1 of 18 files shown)


octobot_trading/exchanges/connectors/ccxt/ccxt_websocket_connector.py
```diff
@@ -632,6 +632,8 @@ async def _wait_for_initialization(self, feed, *g_args, **g_kwargs):
         is_initialized_func = None
         if feed is Feeds.CANDLE:
             def candle_is_initialized_func():
+                if self.exchange_manager is None:
+                    return False
```
Herklos, line 636: Should we raise instead of returning a value?

octobot_trading/personal_data/orders/order.py
```diff
@@ -547,11 +547,13 @@ def update_from_raw(self, raw_order):
...
         stop_price = raw_order.get(enums.ExchangeConstantsOrderColumns.STOP_PRICE.value, None)
-        if stop_price is not None and (price is None or price == 0):
-            # parse stop price when available
-            price = stop_price
+        stop_loss_price = raw_order.get(enums.ExchangeConstantsOrderColumns.STOP_LOSS_PRICE.value, None)
+        take_profit_price = raw_order.get(enums.ExchangeConstantsOrderColumns.TAKE_PROFIT_PRICE.value, None)
+        for potential_price in (stop_loss_price, take_profit_price, stop_price):
+            if potential_price is not None and (price is None or price == 0):
+                # use stop price when available
+                price = potential_price
```
Herklos, line 556: I think we could simplify with price = stop_price or stop_profit_price or... 
And we assign with get() or None

octobot_trading/exchanges/connectors/ccxt/ccxt_websocket_connector.py
GuillaumeDSM: this happens in tests, it's just to avoid raising attribute errror

Herklos (approved):

PR #917, opened 2023-05-10
[Signals] properly handle chained and groupped orders signals

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Commons/pull/367

### DIFF
--- a/octobot_trading/personal_data/orders/orders_manager.py
+++ b/octobot_trading/personal_data/orders/orders_manager.py
@@ -157,6 +157,15 @@ class OrdersManager(util.Initializable):
                 return pending_order
         return None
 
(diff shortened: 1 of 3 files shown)


octobot_trading/personal_data/orders/orders_manager.py
```diff
@@ -157,6 +157,15 @@ def _get_pending_order(self, created_order, should_pop):
                 return pending_order
         return None
 
+    def get_all_active_and_pending_orders_shared_signal_order_id(self) -> list:
+        return [
```
Herklos, line 161: Could we use _select_order instead of adding a new orders.values() iteration method ?

Herklos (approved):

PR #918, opened 2023-05-12
Chained orders signals

GuillaumeDSM:


octobot_trading/personal_data/orders/order.py
```diff
@@ -48,8 +48,8 @@ def __init__(self, trader, side=None):
         self.simulated = trader.simulate
 
         self.logger_name = None
-        self.order_id = trader.parse_order_id(None)
-        self.shared_signal_order_id = str(uuid.uuid4())
+        self.order_id = trader.parse_order_id(None)      # order id: given by the exchange local to the user account
+        self.shared_signal_order_id = str(uuid.uuid4())  # shared order id: kept through instances and trading signals
```
Herklos, line 52: I think we should name it "internal_order_id" or something like this because when I first read about it I expected it to be an order id shared with someone else. As I understand, it could be shared when using copy trading but it may also not be shared. I would prefer to name it differently as it is not always shared with others

GuillaumeDSM: right, i'll change it.
what about `bot_order_id` ? as it will be shared, it's not really internal

octobot_trading/exchanges/adapters/abstract_adapter.py
```diff
@@ -133,6 +134,9 @@ def safe_decimal(self, container, key, default):
         return default
 
     def fix_order(self, raw, **kwargs):
+        # id is reserved for octobot managed id. store exchange id in EXCHANGE_ID
+        raw[enums.ExchangeConstantsOrderColumns.EXCHANGE_ID.value] = \
```
Herklos, line 138: 👍

PR #921, opened 2023-05-15
[Signals] use json dependency for signals strategy config

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Commons/pull/370


Herklos (approved):
👍

PR #931, opened 2023-06-29
[Exchanges] support Binance futures

GuillaumeDSM:
remaining: handle portfolios


octobot_trading/modes/channel/abstract_mode_producer.py
```diff
@@ -441,7 +441,10 @@ async def _register_and_apply_required_user_inputs(self, context):
                 common_enums.ActivationTopics.EVALUATION_CYCLE.value,
                 activation_topic_values
             )
-        await self._apply_exchange_side_config(context)
+        try:
+            await self._apply_exchange_side_config(context)
+        except Exception as err:
+            self.logger.exception(err, True, f"Error when applying exchange side config: {err}")
```
Herklos, line 447: Should we really continue if side config apply failed?

octobot_trading/util/config_util.py
```diff
@@ -63,6 +65,29 @@ def get_symbols(config, enabled_only) -> list:
...
+        if parsed_symbol.is_inverse():
+            return trading_enums.FutureContractType.INVERSE_PERPETUAL.value
+    else:
+        if parsed_symbol.is_linear():
+            return trading_enums.FutureContractType.LINEAR_EXPIRABLE.value
+        if parsed_symbol.is_inverse():
+            return trading_enums.FutureContractType.INVERSE_EXPIRABLE.value
+
+
+def get_symbol_types_counts(config, enabled_only) -> dict:
```
Herklos, line 84: Where is it used?

Herklos: I found it https://github.com/Drakkar-Software/OctoBot-Tentacles/pull/997/files#diff-0facf607e31f87626ab5675ff24e7d4b1fe60449c55c29543b7c7e6e5e60daa8R81 :)

octobot_trading/modes/channel/abstract_mode_producer.py
GuillaumeDSM: For now this is only affecting the leverage setting, if the exchange leverage is different from the configured one. Do you think we should stop the trading mode instead ? (as a note, this error was not caught before and was hidden as this is run in a task)

Herklos: I don't know, I wanted to know your point of view about it. I'm not sure we should continue even if it's only the leverage setting, we are in an unhealthy state.

octobot_trading/modes/channel/abstract_mode_consumer.py
```diff
@@ -131,18 +131,30 @@ async def can_create_order(self, symbol, state):
...
+                f"{portfolio.get_currency_portfolio(currency).available} > {symbol_min_amount}"
+            )
             return portfolio.get_currency_portfolio(currency).available > symbol_min_amount
 
         # long cases => buy => need money(aka other currency in the pair) to buy this currency
         elif state == enums.EvaluatorStates.LONG.value or state == enums.EvaluatorStates.VERY_LONG.value:
+            self.logger.debug(
+                f"can_create_order: portfolio.get_currency_portfolio(market).available > order_min_amount: "
+                f"{portfolio.get_currency_portfolio(market).available} > {order_min_amount}"
+            )
```
Herklos, line 153: Are them WIP logs?

octobot_trading/personal_data/exchange_personal_data.py
```diff
@@ -203,7 +205,11 @@ async def handle_order_update_from_raw(self, exchange_order_id, raw_order,
 
     async def update_order_from_stored_data(self, exchange_order_id, pending_groups):
         order = self.orders_manager.get_order(None, exchange_order_id=exchange_order_id)
+        previous_order_id = order.order_id
         await orders_storage_operations.apply_order_storage_details_if_any(order, self.exchange_manager, pending_groups)
+        if previous_order_id != order.order_id:
+            # order_id got restored to its original value
+            self.orders_manager.replace_order(previous_order_id, order)
```
Herklos, line 212: How the order id can change here?

octobot_trading/personal_data/portfolios/portfolio_manager.py
```diff
@@ -83,13 +83,26 @@ async def handle_balance_update_from_order(self, order, require_exchange_update:
...
+                    portfolio_update_success = await self._refresh_real_trader_portfolio() and portfolio_update_success
+                    # when orders is filled on futures: reload positions to ensure positions sync
+                    if self.exchange_manager.is_future and order.is_filled():
+                        try:
+                            position_instance = order.exchange_manager.exchange_personal_data.positions_manager. \
+                                get_order_position(order)
+                            await self.exchange_manager.exchange_personal_data.positions_manager.\
+                                refresh_real_trader_position(position_instance)
+                        except Exception as err:
+                            self.logger.exception(err)
```
Herklos, line 104: I don't understand why we do it here, It is not related to portfolio.

octobot_trading/personal_data/exchange_personal_data.py
GuillaumeDSM: the initial order id when an order is fetched from exchange is a random uuid generated in constructor, we can't keep it when the order is related to a stored order

octobot_trading/modes/channel/abstract_mode_consumer.py
GuillaumeDSM: Originally they were and I find them very useful to understand why orders are not created, I think we should keep them

octobot_trading/personal_data/exchange_personal_data.py
Herklos: If I understand well, when an order is loaded from the order database we need to restore the order id to match the one in the database right?

GuillaumeDSM: Right, because this order_id is supposed to remain constant for a given order (mainly for trading signals where the order is identified by this id)

Herklos: Right, thanks :+1:

Herklos (approved):
:+1:

PR #959, opened 2023-08-29
[Trades] add exchange_trade_id and aggregator

GuillaumeDSM:


octobot_trading/personal_data/trades/trades_util.py
```diff
@@ -66,3 +66,25 @@ def compute_win_rate(exchange_manager):
...
+            # initialise as a new trade that will contain the aggregation
+            aggregated_trades_by_exchange_order_id[trade.exchange_order_id] = personal_data.create_trade_from_dict(
+                trade.trader, trade.to_dict()
+            )
+        else:
+            # include other trades from the same order id into the aggregated trade
+            to_update = aggregated_trades_by_exchange_order_id[trade.exchange_order_id]
+            to_update.executed_price = (
+                to_update.executed_quantity * to_update.executed_price + trade.executed_quantity * trade.executed_price
+            ) / (to_update.executed_quantity + trade.executed_quantity)
```
Herklos, line 84: I think we should catch some calculation issue here like DivideByZero

GuillaumeDSM: indeed, I updated it, thanks

Herklos (approved):

PR #974, opened 2023-09-24
Trade vol

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Commons/pull/398

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [2.4.28] - 2023-09-24
+### Updated
+- [Trades] push USD-like volume
+
 ## [2.4.27] - 2023-09-12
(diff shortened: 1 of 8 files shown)


octobot_trading/constants.py
```diff
@@ -110,6 +110,13 @@
                               "hitbtc", "kraken", "poloniex", "bitso", "ndax", "upbit",
                               "wavesexchange"]
 
+# from https://www.coingecko.com/en/categories/stablecoins
+USD_LIKE_COINS = [
+    "USDT", "USDC", "TUSD", "BUSD", "DAI", "USD",
+    "FRAX", "USDD", "USDP", "GUSD", "LUSD", "FDUSD",
+    "ALUSD", "USDTC", "MIM", "DOLA", "BOB", "MIMATIC"
+]
```
Herklos, line 118: I think we should add it into constants to make it available for every other modules

GuillaumeDSM: It is already in constants or you mean something else ?

Herklos: Sorry I mean in commons

GuillaumeDSM: good idea, i'll move it there

Herklos (approved):

PR #979, opened 2023-10-03
[OrderState] avoid fill state crash

GuillaumeDSM:


octobot_trading/personal_data/orders/states/fill_order_state.py
```diff
@@ -81,6 +81,8 @@ async def terminate(self):
         `force_close = True` because we know that the order is successfully filled.
         """
         try:
+            self.ensure_not_cleared(self.order)
+
             self.log_event_message(enums.StatesMessages.FILLED)
 
             # call filling actions
```
Herklos, line 88: Isn't it cleared during `on_fill_actions` ?

GuillaumeDSM: It's cleared after the cancel order state. The issue is when a cleared order calls `self.order.generate_executed_time()` (a few lines bellow), it crashes

Herklos: Are we sure the previous `self.ensure_not_cleared(self.order)` is not necessary now?

GuillaumeDSM: Right, in case it's cleared in on_fill_action, i'll add a check

Herklos (approved):

PR #981, opened 2023-10-04
[Signals] add sort_signals

GuillaumeDSM:


octobot_trading/signals/trading_signal_bundle_builder.py
```diff
@@ -32,6 +32,24 @@ def build(self) -> signals.SignalBundle:
         self._pack_referenced_orders_together()
         return super().build()
 
+    def sort_signals(self):
```
Herklos, line 35: I think we could improve this method by using `sort` and a comparator

GuillaumeDSM: I think it would make it very complex for no reason here as we would need to keep the original order of signals but only "moving" cancel order at the begining of the list.
Or maybe you see a way to do it easily and while being sure to keep initial order ?

Herklos: It could be sorted in 5 lines and without iterating 2 times on the array like https://stackoverflow.com/questions/5213033/sort-a-list-of-lists-with-a-custom-compare-function

GuillaumeDSM: udpated

tests/signals/test_trading_signal_bundle_builder.py
```diff
@@ -49,6 +49,34 @@ def test_build(trading_signal_bundle_builder):
         _pack_referenced_orders_together_mock.assert_called_once()
 
 
+def test_sort(trading_signal_bundle_builder, buy_limit_order):
```
Herklos, line 52: :+1:

octobot_trading/signals/trading_signal_bundle_builder.py
Herklos: :+1:

Herklos (approved):

PR #984, opened 2023-10-08
[TradingMode] add INITIAL_PORTFOLIO_OPTIMIZATION

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Commons/pull/404


octobot_trading/modes/channel/abstract_mode_producer.py
```diff
@@ -480,3 +495,19 @@ async def _apply_exchange_side_config(self, context):
                 await util.wait_for_topic_init(self.exchange_manager, self.CONFIG_INIT_TIMEOUT,
                                                common_enums.InitializationEventExchangeTopics.CONTRACTS.value)
             await script_keywords.set_leverage(context, await script_keywords.user_select_leverage(context))
+
+    @classmethod
+    def producer_exchange_wide_lock(cls, exchange_manager) -> asyncio_tools.RLock():
+        try:
+            return cls.PRODUCER_LOCKS_BY_EXCHANGE_ID[exchange_manager.id]
```
Herklos, line 502: :+1:

octobot_trading/personal_data/orders/order_util.py
```diff
@@ -537,3 +537,30 @@ async def get_order_size_portfolio_percent(exchange_manager, order_amount, side,
...
+        if order.state is None:
+            logging.get_logger(LOGGER_NAME).error(
+                f"None state on created order, impossible to wait for fill, order: {order}"
+            )
+        else:
+            try:
+                await order.state.wait_for_next_state(timeout)
+            except asyncio.TimeoutError:
+                logging.get_logger(LOGGER_NAME).error(
+                    f"Timeout while waiting for rebalance marker open order fill, order {order}"
```
Herklos, line 553: It may not be a rebalance order as it is in order_util :smile:

octobot_trading/personal_data/orders/order_util.py
```diff
@@ -537,3 +537,30 @@ async def get_order_size_portfolio_percent(exchange_manager, order_amount, side,
...
+                )
+            if wait_for_portfolio_update and isinstance(order.state, fill_order_state.FillOrderState):
+                # portfolio is updated in FillOrderState: wait for this state to complete
```
Herklos, line 556: It concerns local portfolio update right? Or it fetches the exchange?

octobot_trading/modes/modes_util.py
```diff
@@ -45,3 +53,108 @@ async def clear_plotting_cache(trading_mode):
...
+    created_orders = []
+    for asset in sellable_assets:
+        new_orders, tickers = await convert_asset_to_target_asset(
+            trading_mode, asset, target_asset, asset_amount=None, tickers=tickers
+        )
+        created_orders += new_orders
+    return created_orders, tickers
+
+
+async def convert_asset_to_target_asset(
```
Herklos, line 69: can we split this method in sub methods please, it's a bit hard to read

octobot_trading/modes/modes_util.py
```diff
@@ -45,3 +53,108 @@ async def clear_plotting_cache(trading_mode):
...
+                price = decimal.Decimal(str(
+                    tickers[symbol][trading_enums.ExchangeConstantsTickersColumns.CLOSE.value]
+                    or tickers[symbol][trading_enums.ExchangeConstantsTickersColumns.PREVIOUS_CLOSE.value]
+                ))
+            except KeyError:
+                # can't get price, should not happen as symbol is in client_symbols
+                trading_mode.logger.error(
+                    f"Impossible to convert {asset} into {target_asset}: {symbol} ticker can't be fetched"
+                )
+                return created_orders, tickers
```
Herklos, line 115: I don't understand, we are trying to get the price of a traded symbol here?

octobot_trading/personal_data/orders/order_util.py
GuillaumeDSM: :o right

octobot_trading/personal_data/orders/order_util.py
GuillaumeDSM: both, actually when it's about local pf, this wait is not even done as order is in closed state. Here it's mostly to wait for the portfolio refresh that is triggered when an order is filled. otherwise we proceed with a filled order and out of sync portfolio

octobot_trading/modes/modes_util.py
GuillaumeDSM: if it's a traded symbol, price will be foundin "1. try with converter", however when rebalacing we might need price of symbols we are not trading (if we want to sell USD into USDT and we don't have any /USD trading pair)

Herklos: Yes ok, I was wondering why we have to fetch tickers. Does it exist scenarios where the ticker is not fetch or it will always be fetched ? I'm asking because I think if it's always used we should pass it as a param of the function instead of fetching them inside the function

GuillaumeDSM: ticker is only fetched when trading assets we can't get the price of otherwise (it is passed as a param and only fetched once)

tests/modes/test_abstract_trading_mode.py
```diff
@@ -247,3 +246,58 @@ async def test_edit_order(trading_mode, buy_limit_order):
...
+    with mock.patch.object(
+        modes.AbstractTradingMode, "single_exchange_process_optimize_initial_portfolio", mock.AsyncMock(side_effect=waiter)
+    ) as single_exchange_process_optimize_initial_portfolio_mock, mock.patch.object(
+        exchange_util, "get_common_traded_quote", mock.Mock(return_value="USDT")
+    ) as get_common_traded_quote_mock:
+        mode_1_orders, mode_2_orders = await asyncio.gather(
+            mode_1.optimize_initial_portfolio(["BTC"], {}),
+            mode_2.optimize_initial_portfolio(["BTC"], {}),
+        )
+        # mode_2 did not call get_common_traded_quote_mock nor single_exchange_process_optimize_initial_portfolio_mock
```
Herklos, line 272: :+1:

tests/modes/test_modes_util.py
```diff
@@ -0,0 +1,203 @@
...
+    trading_mode.create_order.assert_called_once()
+    trading_mode.create_order.reset_mock()
+
+    orders = await modes_util.convert_asset_to_target_asset(
+        trading_mode, "USDT", "ETH", tickers, asset_amount=decimal.Decimal(450)
+    )
+    assert len(orders) == 1
+    order = orders[0]
+    assert order.order_type == trading_enums.TraderOrderType.BUY_MARKET
+    assert order.symbol == "ETH/USDT"
```
Herklos, line 192: :+1:

Herklos (approved):

PR #1001, opened 2023-11-04
[Orders] avoid close state crash

GuillaumeDSM:


octobot_trading/personal_data/orders/states/close_order_state.py
```diff
@@ -40,6 +40,9 @@ async def on_refresh_successful(self):
         """
         Verify the order is properly closed
         """
+        if self.order is None:
```
Herklos, line 43: Shouldn't we check if the order is cleared?

GuillaumeDSM: updated

Herklos (approved):

PR #1006, opened 2023-12-08
[TradingMode] init health check

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Commons/pull/412

### DIFF
--- a/octobot_trading/modes/channel/abstract_mode_producer.py
+++ b/octobot_trading/modes/channel/abstract_mode_producer.py
@@ -342,7 +342,7 @@ class AbstractTradingModeProducer(modes_channel.ModeChannelProducer):
         await asyncio.wait_for(self._is_trigger_completed.wait(), timeout=timeout)
 
     @contextlib.asynccontextmanager
-    async def trading_mode_trigger(self):
+    async def trading_mode_trigger(self, skip_health_check=False):
         try:
             self._is_trigger_completed.clear()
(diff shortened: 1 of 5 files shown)


GuillaumeDSM:
> Should we add a param to disable health check? It can be usefull for staggered / grid for example

For DCA, we can by setting the theshold to 100, do you think we need more than this ? agree for the grid part, i'll add it

Herklos:
> > Should we add a param to disable health check? It can be usefull for staggered / grid for example
> 
> For DCA, we can by setting the theshold to 100, do you think we need more than this ? agree for the grid part, i'll add it

I'm not sure we should use an "indirect" value like a 100% threshold to disable healthcheck, it's not easy to understand in my opinion. We may have something like a checkbox (or equivalent) to allow users to enable / disable health check.

octobot_trading/modes/channel/abstract_mode_producer.py
```diff
@@ -354,6 +354,8 @@ async def trading_mode_trigger(self):
                 except asyncio.TimeoutError as e:
                     raise errors.InitializingError() from e
                 self.logger.debug("Order initialized")
+            if self.trading_mode.should_trigger_health_check() and not skip_health_check:
```
Herklos, line 357: :+1:

Herklos (approved):

Herklos (approved):

PR #1025, opened 2024-03-05
Retrier error

GuillaumeDSM:


Herklos (approved):

octobot_trading/exchanges/abstract_exchange.py
```diff
@@ -479,12 +480,18 @@ async def _retry_until(self, timeout, n_times, request_func, *args, **kwargs):
...
+                latest_error = err
+                self.logger.debug(
+                    f"Request retrier failed for {request_func.__name__}({args} {kwargs}) (attempt {attempt}) ({err})"
+                )
                 if time.time() - last_request_time < minimal_interval:
                     await asyncio.sleep(minimal_interval)
                 attempt += 1
-        raise errors.FailedRequest(f"Failed to successfully run {request_func.__name__} request after {attempt} "
-                                   f"attempts.")
+        latest_error = latest_error or RuntimeError("unknown error, this is unexpected, latest_error should be set")
```
Herklos, line 490: Should we rreally use RuntimeError here ? Not simple error ?

GuillaumeDSM: we can, it's transformed directly into a failedrequest at the line just after anyway. You think Exeception is better here ?

Herklos: Yes I don't understand why using a RuntimeError vs a "classic" Exception

GuillaumeDSM: pylint recommends not to raise Exception, that's why I used RuntimeError instead https://pylint.pycqa.org/en/latest/user_guide/messages/warning/broad-exception-raised.html

Now it has no consequence here, we could do both

PR #1028, opened 2024-03-11
Fees

GuillaumeDSM:


tests/modes/test_modes_util.py
```diff
@@ -195,6 +195,54 @@ async def test_convert_asset_to_target_asset(backtesting_trader):
...
+
+        # cast 2: reduce amount to cover fees
+        orders = await modes_util.convert_asset_to_target_asset(
+            trading_mode, "USDT", "ETH", tickers, asset_amount=decimal.Decimal(1000)
+        )
+        assert len(orders) == 1
+        order = orders[0]
+        assert order.order_type == trading_enums.TraderOrderType.BUY_MARKET
+        assert order.symbol == "ETH/USDT"
+        assert order.origin_quantity == decimal.Decimal("0.66400000")   # lower than 0.66666666 when fees is 0 USDT
```
Herklos, line 240: :+1:

octobot_trading/modes/modes_util.py
```diff
@@ -151,19 +151,29 @@ def _get_associated_symbol_and_order_type(trading_mode, asset: str, target_asset
...
-            quantity = asset_amount if order_type is trading_enums.TraderOrderType.SELL_MARKET \
-                else (asset_amount / price)
+            quantity = asset_amount if side is trading_enums.TradeOrderSide.SELL else (asset_amount / price)
         except (decimal.DivisionByZero, decimal.InvalidOperation):
             quantity = constants.ZERO
-    return quantity
+
+    adapted_quantity = trading_personal_data.decimal_adapt_order_quantity_because_fees(
+        trading_mode.exchange_manager, symbol, order_type, quantity, price,
+        trading_enums.ExchangeConstantsMarketPropertyColumns.TAKER, side,
```
Herklos, line 173: Here we use TAKER because we expect to be taker ?

GuillaumeDSM: yes, because we can't be sure that a limit order will be taken as maker so we prefer being pessimistic and expect taker fees

Herklos: I see, seems legit

Herklos (approved):

PR #1040, opened 2024-03-21
Index

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Commons/pull/423


octobot_trading/modes/channel/abstract_mode_consumer.py
```diff
@@ -69,17 +69,24 @@ async def internal_callback(self, trading_mode_name, cryptocurrency, symbol, tim
...
-            }
-        except Exception as err:
-            self.logger.error(f"Error when getting funds for {symbol}: {err}")
-            funds = {}
+        if symbol is None:
+            return (
+                f"Not enough funds to create new orders after {final_note} evaluation: "
+                f"{self.exchange_manager.exchange_name} exchange minimal order "
+                f"volume has not been reached."
+            )
```
Herklos, line 77: That's the reason why symbol is None?

tests/modes/test_abstract_mode_consumer.py
```diff
@@ -169,6 +173,70 @@ async def test_get_holdings_ratio():
...
+    assert consumer.get_holdings_ratio("ETH", include_assets_in_open_orders=True) \
+           == decimal.Decimal('0.01818181818181818181818181818')
+    assert consumer.get_holdings_ratio("USDT", include_assets_in_open_orders=True) \
+           == decimal.Decimal('0.09090909090909090909090909091')
+
+    await exchange_manager.exchange_personal_data.orders_manager.upsert_order_instance(order_3)
+    assert consumer.get_holdings_ratio("BTC", include_assets_in_open_orders=True) \
+           == decimal.Decimal('1')
+    assert consumer.get_holdings_ratio("ETH", include_assets_in_open_orders=True) \
+           == decimal.Decimal('0.01819090909090909090909090909')
```
Herklos, line 235: I don't understand why this value doesn't change due to order_3 addition

octobot_trading/modes/channel/abstract_mode_consumer.py
GuillaumeDSM: in index trading, we don't send a given symbol to buy, we tell the consumer to "rebalance the index" and the consumer will create orders accordingly (it's not symbol related).

Currently:
- producer decides if a rebalance should happen
- consumer executes the rebalance

tests/modes/test_abstract_mode_consumer.py
GuillaumeDSM: we are selling XRP for ETH in order3 making more ETH available when it will be filled

Herklos: So why the ETH ratio is the same as without order_3 ?

octobot_trading/modes/channel/abstract_mode_consumer.py
Herklos: >  (it's not symbol related).

That's my point, I don't understand why we trigger this error here when symbol is None

GuillaumeDSM: This function's goal is to log an error, if it's called it's that the error already occured and we want inform the user that an error occured. This symbol condition allows to create the error without given symbols, which is a case that can now happen

tests/modes/test_abstract_mode_consumer.py
GuillaumeDSM: it's not the same, before it was `0.01818181818181818181818181818` and after it's `0.01819090909090909090909090909`

Herklos: Oooh right, my bad. It saw the same number :smile:

Herklos (approved):
:+1:

Herklos (approved):

PR #1077, opened 2024-07-23
[Orders] add disabled_order_auto_synchronization

GuillaumeDSM:


octobot_trading/personal_data/state.py
```diff
@@ -236,6 +239,9 @@ async def on_refresh_successful(self):
         """
         raise NotImplementedError("on_refresh_successful not implemented")
 
+    def _is_synchronization_enabled(self):
+        return True
```
Herklos, line 243: is it overwritten somewhere?

octobot_trading/util/test_tools/exchanges_test_tools.py
```diff
@@ -245,5 +253,7 @@ async def wait_for_other_status(order: personal_data.Order, timeout) -> personal
             return personal_data.create_order_instance_from_raw(
                 order.trader, raw_order, force_open_or_pending_creation=False
             )
+        if time.time() - t0 + constants.CREATED_ORDER_FORCED_UPDATE_PERIOD >= timeout:
+            break
```
Herklos, line 257: should we add a debug log here or it's not necessary?

octobot_trading/personal_data/state.py
GuillaumeDSM: yes, in OrderState

octobot_trading/util/test_tools/exchanges_test_tools.py
GuillaumeDSM: No we don't this check is just to kip an asyncio.sleep when we know the next while condition will be False (and therefore we sleep for nothing)

octobot_trading/personal_data/state.py
Herklos: Right, my bad!

Herklos (approved):
👍

PR #1082, opened 2024-08-10
MM

GuillaumeDSM:


octobot_trading/exchanges/traders/trader.py
```diff
@@ -459,6 +459,29 @@ async def _handle_order_cancel_error(self, order, err, wait_for_cancelling, canc
...
+        try:
+            cancel_on_exchange = False
+            for order in orders_to_cancel:
+                await order.lock.acquire()
+                if self.simulate or order.is_self_managed():
+                    order.status = octobot_trading.enums.OrderStatus.CANCELED
+                else:
+                    cancel_on_exchange = True
+            if cancel_on_exchange:
+                await self.exchange_manager.exchange.cancel_all_orders(symbol)
```
Herklos, line 476: Should we add a fallback mechanism if cancel all orders is not supported ?

GuillaumeDSM: good question. I don't really know if we should as the fallback would be to cancel all orders one by one and i'm not sure we want that when calling this method.

Maybe we can add a param to allow fallback or not

Herklos: Yes we could add a param to allow fallback. If it had been supported on most exchanges, we could perhaps have left it as is, but it doesn't seem to be the case and so if there is no fallback it is not very logical compared to our framework which abstracts the specificities of exchanges.

GuillaumeDSM: I'll add it 👍

GuillaumeDSM: up

octobot_trading/exchanges/traders/trader.py
```diff
@@ -459,6 +459,53 @@ async def _handle_order_cancel_error(self, order, err, wait_for_cancelling, canc
                 f"Can't cancel order and unknown post sync order state for order: {order}."
             ) from err
 
+    async def cancel_all_orders(
+        self,
+        symbol: str,
+        allow_single_order_cancel_fallback: bool,
```
Herklos, line 465: 👍

octobot_trading/exchanges/traders/trader.py
```diff
@@ -459,6 +459,53 @@ async def _handle_order_cancel_error(self, order, err, wait_for_cancelling, canc
...
+                    success = False
+                    await self.exchange_manager.exchange.cancel_all_orders(symbol)
+                    success = True
+                except errors.NotSupported:
+                    if allow_single_order_cancel_fallback:
+                        self.logger.debug(
+                            f"cancel_all_orders is not supported on {self.exchange_manager.exchange_name}. Falling "
+                            f"back to one by one cancel"
+                        )
+                        for order in orders_to_cancel:
```
Herklos, line 490: 👍

Herklos (approved):

Herklos (approved):

Herklos (approved):

PR #1104, opened 2024-09-05
Exchange fixes

GuillaumeDSM:


octobot_trading/modes/channel/abstract_mode_consumer.py
```diff
@@ -198,6 +198,10 @@ async def can_create_order(self, symbol, state):
             )
             return can_create_order
 
+        elif state == enums.EvaluatorStates.NEUTRAL.value:
+            self.logger.debug(f"can_create_order: True. No check to perform for {state} state")
```
Herklos, line 202: Shouldn't it be `can_create_order: {can_create_order}`?

GuillaumeDSM: Not really because the `can_create_order` variable doesn't exist in this branch of the code; Now we could add it if you think it makes it clearer. The idea of this log is just to track the fact that the order signal didn't get filtered out

Herklos: I see, thanks for these details

Herklos (approved):

Herklos (approved):

PR #1127, opened 2024-11-17
[TradingMode] fix convertor with simulated liimit order

GuillaumeDSM:


octobot_trading/modes/modes_util.py
```diff
@@ -142,7 +142,14 @@ async def convert_asset_to_target_asset(
                 quantity=order_quantity,
                 price=order_price
             )
-            created_orders.append(await trading_mode.create_order(order))
+            initialized_order = await trading_mode.create_order(order)
+            if (
+                isinstance(initialized_order, trading_personal_data.LimitOrder)
+                and initialized_order.trader.exchange_manager.trader.simulate
```
Herklos, line 148:
```suggestion
                and initialized_order.is_simulated()
```
I don't remember if this method exists

GuillaumeDSM: right, there is a `simulated` attribute, I update this

GuillaumeDSM: up

octobot_trading/modes/modes_util.py
```diff
@@ -142,7 +142,11 @@ async def convert_asset_to_target_asset(
                 quantity=order_quantity,
                 price=order_price
             )
-            created_orders.append(await trading_mode.create_order(order))
+            initialized_order = await trading_mode.create_order(order)
+            if isinstance(initialized_order, trading_personal_data.LimitOrder) and initialized_order.simulated:
+                # on simulator, this order should be instantly filled now as its price is meant to be instantly filled
+                await initialized_order.on_fill()
```
Herklos, line 148: Shouldn't we create a wrapper like `create_market_order` that creates a market order or a limit order (depending on what's available for this exchange) and that triggers order.on_fill() for the limit order. 
It seems to be an issue that could happen again in another function who would do the same thing

GuillaumeDSM: right, it's now in `convert_with_market_or_limit_order`

Herklos (approved):
👍

PR #1130, opened 2024-11-23
handle historical config

GuillaumeDSM:
requires https://github.com/Drakkar-Software/OctoBot-Commons/pull/449

### DIFF
--- a/octobot_trading/modes/abstract_trading_mode.py
+++ b/octobot_trading/modes/abstract_trading_mode.py
@@ -16,11 +16,13 @@
 import abc
 import contextlib
 import decimal
+import copy
 
 import octobot_commons.constants as common_constants
 import octobot_commons.enums as common_enums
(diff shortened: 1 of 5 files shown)


octobot_trading/modes/abstract_trading_mode.py
```diff
@@ -432,6 +437,33 @@ async def reload_config(self, bot_id: str, trading_config=None) -> None:
     def get_local_config(self):
         return self.trading_config
 
+    def update_config_and_user_inputs_if_necessary(self):
+        if self.update_config_if_necessary():
+            import octobot_commons.timestamp_util
```
Herklos, line 442: Why importing it here?

GuillaumeDSM: 😮 thanks, that was for a debug print

Herklos (approved):

PR #1131, opened 2024-11-24
Backtesting flex

GuillaumeDSM:


octobot_trading/exchanges/config/exchange_config_data.py
```diff
@@ -114,27 +114,30 @@ def initialize_exchange_event_tree(self):
                     commons_enums.InitializationEventExchangeTopics.ORDER_BOOK,
                     commons_enums.InitializationEventExchangeTopics.FUNDING,
             ):
-                for symbol in self.traded_symbol_pairs:
-                    if topic in (
+                if not self.exchange_manager.is_backtesting:
+                    # disabled in backtesting: not all traded pairs are always available from the start:
+                    # don't wait for them in events
```
Herklos, line 119: I don't understand why we should not create these events. I think we should remove the "await event" in backtesting, not the event tree creation

GuillaumeDSM: If we do that, it means that we will need to add a "if backtesting" everytime we want to use any of those events

GuillaumeDSM: up!

octobot_trading/modes/channel/abstract_mode_producer.py
```diff
@@ -455,20 +455,25 @@ async def _wait_for_symbol_init(self, symbol, time_frame, timeout) -> bool:
             self.logger.error(f"Initialization took more than {timeout} seconds")
         return False
 
-    async def _wait_for_bot_init(self, timeout, extra_topics: list=None) -> bool:
+    async def _wait_for_bot_init(self, timeout, extra_symbol_topics: list=None) -> bool:
         try:
-            topics = [
-                common_enums.InitializationEventExchangeTopics.BALANCE.value,
```
Herklos, line 461: I am not able to find BALANCE in the ouput diff, does BALANCE topic is removed?

GuillaumeDSM: I moved it out of the symbol  for loop

Herklos: Oh yes my bad

Herklos (approved):
:+1:

Herklos (approved):

PR #1137, opened 2024-12-03
2 Fixes

GuillaumeDSM:


octobot_trading/constants.py
```diff
@@ -224,6 +224,10 @@
     FUNDING_CHANNEL: [enums.WebsocketFeeds.FUNDING]
 }
 
+ALWAYS_STARTED_REST_PRODUCER_CHANNELS = [
+    TICKER_CHANNEL, # use to force mark price update when necessary: should always be reachable
```
GuillaumeDSM, line 228: issue when asking for mark price updater (when it is missing) and a ws is active: no ticker updater is available, ending in "IndexError: list index out of range" and no way to force mark price update

Herklos (approved):

PR #1156, opened 2025-01-08
Trigger above

GuillaumeDSM:


octobot_trading/personal_data/orders/order_factory.py
```diff
@@ -105,6 +106,8 @@ def create_order_instance(
         exchange_creation_params=exchange_creation_params,
         associated_entry_id=associated_entry_id
     )
+    if trigger_above is not None and isinstance(order, personal_data.LimitOrder):
+        order.trigger_above = trigger_above
```
Herklos, line 110: I don't think it's the proper place do to this. Why not providing it with order.update and then ignoring this data when it's not necessary?

octobot_trading/personal_data/orders/types/limit/limit_order.py
```diff
@@ -31,12 +31,22 @@ def __init__(self, trader, side=enums.TradeOrderSide.BUY):
...
             self.trigger_above = raw_order[enums.ExchangeConstantsOrderColumns.TRIGGER_ABOVE.value]
         return updated
 
+    def to_dict(self):
+        return {
+            **super().to_dict(),
+            **{
+                enums.ExchangeConstantsOrderColumns.TRIGGER_ABOVE.value: self.trigger_above,
+            }
+        }
```
Herklos, line 48: Why do we have to do this?

octobot_trading/signals/util.py
```diff
@@ -36,6 +37,9 @@ def create_order_signal_content(
         trading_enums.TradingSignalOrdersAttrs.EXCHANGE.value: exchange_manager.exchange_name,
         trading_enums.TradingSignalOrdersAttrs.EXCHANGE_TYPE.value: exchanges.get_exchange_type(exchange_manager).value,
         trading_enums.TradingSignalOrdersAttrs.SIDE.value: order.side.value,
+        trading_enums.TradingSignalOrdersAttrs.TRIGGER_ABOVE.value: (
+            order.trigger_above if isinstance(order, personal_data.LimitOrder) else None
```
Herklos, line 41: same here, I don't think this is the proper place to do this check

GuillaumeDSM:
I added trigger_above to Order, it's much cleaner now

Herklos (approved):
:+1:

PR #1162, opened 2025-01-10
[Orders] avoid portfolio desync after filled order

GuillaumeDSM:


GuillaumeDSM:
up with 20s and 1 attempt

octobot_trading/personal_data/portfolios/portfolio_manager.py
```diff
@@ -91,9 +97,43 @@ async def handle_balance_update_from_order(self, order, require_exchange_update:
...
         return False
 
+    async def _refresh_real_trader_portfolio_until_order_update_applied(self, order, expect_filled_order_update):
+        t0 = time.time()
+        refreshed = await self._refresh_real_trader_portfolio()
+        if not expect_filled_order_update:
+            return refreshed
+        if self._has_filled_order_been_applied(order):
+            return refreshed
+        for i in range(constants.MAX_PORTFOLIO_SYNC_ATTEMPTS):
```
Herklos, line 112: Is there a reason to use a loop here as it's only 1 loop?

octobot_trading/constants.py
```diff
@@ -46,6 +46,10 @@
 INSTANT_FILLED_LIMIT_ORDER_PRICE_DELTA = decimal.Decimal("0.005")
 CREATED_ORDER_FORCED_UPDATE_PERIOD = 5
 
+# Portfolio
+MAX_PORTFOLIO_SYNC_ATTEMPTS = 1
+SYNC_ATTEMPTS_INTERVAL = 20
```
Herklos, line 51: 👍

Herklos (approved):

PR #1192, opened 2025-03-05
Stop fix

GuillaumeDSM:


octobot_trading/personal_data/exchange_personal_data.py
```diff
@@ -209,6 +211,19 @@ async def handle_order_update_from_raw(self, exchange_order_id, raw_order,
                 self.logger.exception(e, True, f"Failed to update order : {e}")
         return False, None
 
+    async def _ensure_canceled_out_of_sync_order(self, exchange_order_id: str):
```
Herklos, line 214: Does it has to be async?

GuillaumeDSM: good point, it's up

Herklos (approved):

PR #1197, opened 2025-03-30
Taker limit

GuillaumeDSM:


octobot_trading/personal_data/orders/order.py
```diff
@@ -866,6 +869,18 @@ def _update_type_from_raw(self, raw_order):
             self.exchange_order_type = enums.TradeOrderType.UNKNOWN
         self.side, self.order_type = parse_order_type(raw_order)
 
+
+    def _should_instant_fill(self):
+        if self.order_type in (enums.TraderOrderType.SELL_MARKET, enums.TraderOrderType.BUY_MARKET):
+            return True
```
Herklos, line 875: Why does this code is not in market_order class?

octobot_trading/personal_data/orders/order.py
```diff
@@ -866,6 +869,18 @@ def _update_type_from_raw(self, raw_order):
...
+        if self.origin_price and self.created_last_price:
+            if self.trigger_above:
+                # instant fill if order price is lower or equal to market price: ex spot sell order
+                return self.origin_price <= self.created_last_price
+            # instant fill if order price is higher or equal to market price: ex spot buy order
+            return self.origin_price >= self.created_last_price
+        return False
```
Herklos, line 882: Are you sure it's valid for all orders? Or we consider it will be overwritten by the other Order classes?

octobot_trading/personal_data/orders/order.py
GuillaumeDSM: right, i'll change it

GuillaumeDSM:
it's up!

Herklos (approved):
:+1:

PR #1206, opened 2025-04-25
Inactive orders

GuillaumeDSM:


octobot_trading/exchanges/traders/trader.py
```diff
@@ -385,6 +401,36 @@ async def chain_order(self, order, chained_order, update_with_triggering_order_f
         order.add_chained_order(chained_order)
         self.logger.info(f"Added chained order [{chained_order}] to [{order}] order.")
 
+    async def update_order_as_inactive(
+        self, order, ignored_order=None, wait_for_cancelling=True,
+        cancelling_timeout=octobot_trading.constants.INDIVIDUAL_ORDER_SYNC_TIMEOUT
+    ) -> bool:
+        if self.simulate:
+            self.logger.error(f"Can't update order as inactive on simulated trading.")
+            return False
```
Herklos, line 410: Shouldn't we check ` self.enable_inactive_orders` instead `self.simulate` here ?

octobot_trading/personal_data/orders/order.py
```diff
@@ -95,6 +95,17 @@ def __init__(self, trader, side=None):
...
+        # order activity
+        self.is_active = True   # When is_active=False order is not pushed to exchanges
+        self.active_trigger_price: decimal.Decimal = None    # price threshold from which the order becomes active
+        # when True, order becomes active when current price >= active_trigger_price
+        self.active_trigger_above: bool = None
+        self._active_trigger_event: asyncio.Event = None    # will be set when the price is hit
+        # waiter that will call on_active_trigger() when active_trigger_event is set
+        self._active_trigger_task: asyncio.Task = None
+        # True when a transition between active and inactive is being made
+        self.is_in_active_inactive_transition = False
```
Herklos, line 107: I think we should create an ActiveState or something like this to manage all attributes and functions related to order activeness. I think it will be easier to abstract the activeness logic inside a dedicated class.

octobot_trading/personal_data/orders/order_group.py
```diff
@@ -41,6 +55,24 @@ async def on_cancel(self, cancelled_order, ignored_orders=None):
...
+    async def lock_group(self):
+        if self._lock is None:
+            self._lock = asyncio_tools.RLock()
+        async with self._lock:
+            yield
+
+    def _default_active_order_swap_strategy(self, timeout: float) -> active_order_swap_strategies.ActiveOrderSwapStrategy:
+        """
+        Called when an order of this group is becoming active
+        """
```
Herklos, line 73: Shouldn't it return something ?

Herklos (commented):
Great implementation!

octobot_trading/exchanges/traders/trader.py
GuillaumeDSM: indeed!

octobot_trading/personal_data/orders/order_group.py
GuillaumeDSM: right, thanks 👍

octobot_trading/personal_data/orders/order.py
GuillaumeDSM: done in `order.active_trigger` using the new `PriceTrigger` class

octobot_trading/personal_data/orders/order.py
```diff
@@ -398,13 +426,105 @@ def get_position_side(self, future_contract):
...
+            raise ValueError(
+                f"Both active_trigger_price and active_trigger_above must be provided to set an order as inactive"
+            )
+        logging.get_logger(self.get_logger_name()).info("Order is switching to inactive")
+        self.is_active = False
+        self.create_or_update_active_trigger(active_trigger_price, active_trigger_above)
+        # enforce attributes in case order has been canceled
+        self.status = enums.OrderStatus.OPEN
+        self.canceled_time = 0
+        await self._ensure_inactive_order_watcher()
```
Herklos, line 461: I don't think this is consistent with the trigger architecture. I believe Order should only be able to add/update a trigger but not create it. In other words, it should not take an 'active_trigger_price' argument but only an 'active_trigger,' and it’s the responsibility of the caller to provide the instance of the correct trigger class based on what is desired.

octobot_trading/personal_data/orders/triggers/price_trigger.py
```diff
@@ -0,0 +1,54 @@
...
+#
+#  You should have received a copy of the GNU Lesser General Public
+#  License along with this library.
+import decimal
+import typing
+
+import octobot_trading.personal_data.orders.triggers.base_trigger as base_trigger
+
+
+class PriceTrigger(base_trigger.BaseTrigger):
```
Herklos, line 22: 👍

octobot_trading/personal_data/orders/order.py
GuillaumeDSM: good point, i'll update this, thanks !

GuillaumeDSM: it's up

octobot_trading/personal_data/orders/order.py
```diff
@@ -398,13 +419,108 @@ def get_position_side(self, future_contract):
...
+    @contextlib.contextmanager
+    def active_or_inactive_transition(self):
+        previous_value = self.is_in_active_inactive_transition
+        try:
+            self.is_in_active_inactive_transition = True
+            yield
+        finally:
+            self.is_in_active_inactive_transition = previous_value
+
+    def use_active_trigger(self, active_trigger: price_trigger.PriceTrigger):
```
Herklos, line 431:
```suggestion
    def use_active_trigger(self, active_trigger: base_trigger.BaseTrigger):
```

octobot_trading/personal_data/orders/order.py
```diff
@@ -398,13 +419,108 @@ def get_position_side(self, future_contract):
...
+                f"The current active trigger (at {self.active_trigger.trigger_price}) is still pending, canceling it "
+                f"and replacing it by this new one, this is works but is unexpected."
+            )
+            self.active_trigger.clear()
+            self.active_trigger = active_trigger
+        else:
+            self.active_trigger.trigger_price = active_trigger.trigger_price
+            self.active_trigger.trigger_above = active_trigger.trigger_above
+
+    async def set_as_inactive(self, active_trigger: price_trigger.PriceTrigger):
```
Herklos, line 447:
```suggestion
    async def set_as_inactive(self, active_trigger: base_trigger.BaseTrigger):
```

GuillaumeDSM: indeed, it's up, I also added unit tests.

Herklos (commented):
Great feature!

octobot_trading/personal_data/exchange_personal_data.py
```diff
@@ -298,13 +298,20 @@ async def handle_closed_order_update(self, exchange_order_id, raw_order) -> bool
...
-        for order in self.orders_manager.get_all_orders(symbol=symbol, active=False):
+    ) -> int:
+        activated_orders_count = 0
+        sorted_inactive_orders = sorted(
+            self.orders_manager.get_all_orders(symbol=symbol, active=False),
+            key= lambda o: o.origin_price if o.trigger_above else -o.origin_price
+        )
+        for order in sorted_inactive_orders:
             if order.should_become_active(price_time, current_price):
+                activated_orders_count += 1
```
Herklos, line 309: shouldn't it be after the `on_active_trigger` call?

GuillaumeDSM: Even if the next line fails, we want to know that something might have happened. I'll rename this variable, it’s not really how many got activated but more how many got changed

Herklos: understood

Herklos (approved):

